Issue #7, opened 2017-12-19
Add trait that allows using `as` operator for generic code


cuviper:
_From @bluss on March 23, 2016 21:26_

The `as` operator does unchecked conversion between numerical types, without error cases. It's a core part of the language, so it would be useful to have it available for generic code as well.

Often it can be used very sensibly because you know your input is well behaved. For example, it can be small magnitude floating point numbers that you want to cast to/from f32/f64.

_Copied from original issue: rust-num/num#183_

cuviper:
_From @Binero on March 29, 2016 21:27_

It also appears impossible to cast from `f32` or `f64` to `num::traits::Float`.

cuviper:
_From @bluss on March 29, 2016 21:33_

Yes, `Float` is not a "trait object compatible" trait (a.k.a. "object safe"), so it can not be used as a type like that. I don't think that's expected, not every trait is made to be object compatible.

cuviper:
_From @Binero on March 31, 2016 13:8_

It does make code generic over Float very hard to write, as it's not possible to use constants. In my specific instance, I had to multiply the `Float` input by Euler's number, which was impossible. I cannot make a `Float` from an `f64`.

cuviper:
_From @bluss on March 31, 2016 13:30_

Euler's number is accessible from `Float` though, `F::exp(F::one()) where F: Float`. Not so nice maybe..

cuviper:
_From @bluss on March 31, 2016 13:32_

I'm using [asprim](https://crates.io/crates/asprim) in the mean time, but it is much better for crate interoperation if this kind of feature is already included in the float trait.

cuviper:
_From @hauleth on July 15, 2016 9:55_

You can always use `FromPrimitive` or `From`.

cuviper:
_From @Enet4 on November 2, 2017 11:12_

Just bumping this issue, as I believe that this crate could benefit from an `AsPrim` trait. I have a use case where an N-dimensional volume of primitive elements are linearly transformed and converted to a compile-time value type `T` on reading. The other traits don't work because I would prefer precision loss over not obtaining a value at all. The `asprim` crate is nice, but this being covered by `num` is almost a normal expectation, and would more easily implement the trait for other numeric types such as `Complex`.

cuviper:
> more easily implement the trait for other numeric types such as Complex.

[...]

PR #17, opened 2017-12-20
Add AsPrimitive trait for generic casting with `as`

Enet4:
This is my personal attempt at #7. It is fairly similar to what can be found in `asprim`, although implemented from scratch. Please let me know of what you think. Could it use more tests? Should I also leave a safety notice that some conversions with `as` are currently UB (rust-lang/rust#10184)?

### DIFF
--- a/src/cast.rs
+++ b/src/cast.rs
@@ -462,6 +462,28 @@ impl<T: NumCast> NumCast for Wrapping<T> {
 /// let three: i32 = (3.14159265f32).as_();
 /// assert_eq!(three, 3);
 /// ```
+/// 
+/// # Safety
+/// 
+/// Currently, some uses of the `as` operator are not entirely safe.
+/// In particular, it is undefined behavior if:
+/// 
+/// - A truncated floating point value cannot fit in the target integer
+///   type ([#10184](https://github.com/rust-lang/rust/issues/10184));
+/// 
+/// ```ignore
+/// # use num_traits::AsPrimitive;
+/// let x: u8 = (1.04E+17).as_(); // UB
+/// ```
+/// 
+/// - Or a floating point value does not fit in another floating
+///   point type ([#15536](https://github.com/rust-lang/rust/issues/15536)).
+///
+/// ```ignore
+/// # use num_traits::AsPrimitive;
+/// let x: f32 = (1e300f64).as_(); // UB
+/// ```
+/// 
 pub trait AsPrimitive<T>: 'static + Copy
 where
     T: 'static + Copy
--- a/src/cast.rs
+++ b/src/cast.rs
@@ -493,6 +493,7 @@ impl_as_primitive!(isize => u8, i8, u16, i16, u32, i32, u64, isize, usize, i64, 
 impl_as_primitive!(f32 => u8, i8, u16, i16, u32, i32, u64, isize, usize, i64, f32, f64);
 impl_as_primitive!(f64 => u8, i8, u16, i16, u32, i32, u64, isize, usize, i64, f32, f64);
 impl_as_primitive!(char => char, u8, i8, u16, i16, u32, i32, u64, isize, usize, i64);
+impl_as_primitive!(bool => u8, i8, u16, i16, u32, i32, u64, isize, usize, i64);
 
 #[test]
 fn to_primitive_float() {
--- a/src/cast.rs
+++ b/src/cast.rs
@@ -452,6 +452,48 @@ impl<T: NumCast> NumCast for Wrapping<T> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/cast.rs
```diff
@@ -452,6 +452,48 @@ impl<T: NumCast> NumCast for Wrapping<T> {
...
+impl_as_primitive!(i16 => u8, i8, u16, i16, u32, i32, u64, isize, usize, i64, f32, f64);
+impl_as_primitive!(u32 => u8, i8, u16, i16, u32, i32, u64, isize, usize, i64, f32, f64);
+impl_as_primitive!(i32 => u8, i8, u16, i16, u32, i32, u64, isize, usize, i64, f32, f64);
+impl_as_primitive!(u64 => u8, i8, u16, i16, u32, i32, u64, isize, usize, i64, f32, f64);
+impl_as_primitive!(i64 => u8, i8, u16, i16, u32, i32, u64, isize, usize, i64, f32, f64);
+impl_as_primitive!(usize => u8, i8, u16, i16, u32, i32, u64, isize, usize, i64, f32, f64);
+impl_as_primitive!(isize => u8, i8, u16, i16, u32, i32, u64, isize, usize, i64, f32, f64);
+impl_as_primitive!(f32 => u8, i8, u16, i16, u32, i32, u64, isize, usize, i64, f32, f64);
+impl_as_primitive!(f64 => u8, i8, u16, i16, u32, i32, u64, isize, usize, i64, f32, f64);
+impl_as_primitive!(char => char, u8, i8, u16, i16, u32, i32, u64, isize, usize, i64);
```
cuviper: Interesting that you include `char` -- I think we could justify `bool` to integers as well.  Pointers can also use `as`, but I think it's fair to leave that outside the realm of `num`.

Enet4:
I agree with the conversion from `bool` to integer types, just punched them in.

I might also add that this implementation has this tiny catch (does `asprim` have this as well?):

```rust
let pi: f64 = 3.1415926535897932384626433832795;
if pi.as_() == 3_u16 {
    println!("All good");
}
```

The compiler cannot resolve to `AsPrimitive<u16>` here, because there could be another `AsPrimitive<T>` where `T: PartialEq<u16>`. Making it explicit or performing the comparison the other way around works.

```rust
let pi: f64 = 3.1415926535897932384626433832795;
if 3_u16 == pi.as_() {
    println!("All good");
}
```

Enet4:
Just a heads-up: I forgot to add a Safety section the other day, so I added it now. What do you think?

I configured the new doctests in the description to be ignored, although they compile fine. I'm not sure whether testing whether it's important to compile some UB-inducing code.

bluss:
Lgtm. Give it a thought -- is this an interface that others can implement, should they, or should they not? Documenting that rule would be good. Most probable rule is that the impl is only "admitted" if there is an corresponding `as` cast.

bluss:
As always, the best method for implementation isn't necessarily the best interface for calling. Trait into incidentally also has this problem. What I think of as the best interface for calling is something that admits `foo.as_::<NewType>()`

Enet4:
Thanks for the feedback, @bluss.

> Give it a thought -- is this an interface that others can implement, should they, or should they not? Documenting that rule would be good. Most probable rule is that the impl is only "admitted" if there is an corresponding as cast.

Agreed. It's hard to imagine good use cases for external implementations, though. Would it be for primitive newtypes (say, `struct Foo(u32)`)? I wonder if expressions such as `(5.0).as_().into()` can be resolved unambiguously.

> What I think of as the best interface for calling is something that admits foo.as_::<NewType>(). AsPrim has that, incidentally, but it has the other drawbacks of that.

Yep, I am aware of this issue. This sounds like a blue-red pill decision between having more possible conversion sources/targets (`char` and `bool`) or having a better user-facing API. Whichever is more appealing to the community, I'll make the necessary changes if need be.

On the other hand, given that `asprim` seems to be well maintained right now, choosing to keep this API here would give users the power to choose either one as they wish (or both, but that would be pointless and just terrible 🙃).

cuviper:
The `noisy_float` crate comes to mind as an example that might implement this.

cuviper:
I think I'm ready to merge -- are you happy with those final questions of the interface as-is?

Enet4:
My apologies, I sort of forgot this PR. In the following hours I could add a statement to the documentation mentioning that external types may implement `AsPrimitive` when they behave like a primitive numeric type, _and_ the intended conversion does not fail (although it can result in precision loss or truncation). This should enable types such as those found in `noisy_float` to implement the trait without looking unintuitive.

cuviper:
Looks good, thanks!

bors r+

PR #21, opened 2018-01-03
Add checked shifts

fabianschuiki:
Add traits `CheckedShl` and `CheckedShr` that correspond to the standard
library's `checked_shl` and `checked_shr` functions. Implement the trait
on all primitive integer types by default, akin to what the standard
library does.

The stdlib is somewhat inconsistent when it comes to the type of the
shift amount. The `checked_*` functions have a `u32` shift amount, but
the `std::ops::{Shl,Shr}` traits are generic over the shift amount. Also
the stdlib implements these traits for all primitive integer types as
right-hand sides. Our implementation mimics this behaviour.

### DIFF
--- a/src/int.rs
+++ b/src/int.rs
@@ -21,6 +21,8 @@ pub trait PrimInt
     + CheckedSub<Output=Self>
     + CheckedMul<Output=Self>
     + CheckedDiv<Output=Self>
+    + CheckedShl<usize, Output=Self>
+    + CheckedShr<usize, Output=Self>
     + Saturating
 {
     /// Returns the number of ones in the binary representation of `self`.
--- a/src/int.rs
+++ b/src/int.rs
@@ -21,8 +21,6 @@ pub trait PrimInt
     + CheckedSub<Output=Self>
     + CheckedMul<Output=Self>
     + CheckedDiv<Output=Self>
-    + CheckedShl<usize, Output=Self>
-    + CheckedShr<usize, Output=Self>
     + Saturating
 {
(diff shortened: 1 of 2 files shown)


src/int.rs
```diff
@@ -21,6 +21,8 @@ pub trait PrimInt
     + CheckedSub<Output=Self>
     + CheckedMul<Output=Self>
     + CheckedDiv<Output=Self>
+    + CheckedShl<usize, Output=Self>
+    + CheckedShr<usize, Output=Self>
```
cuviper: Unfortunately, this is a breaking change.  We don't control all implementations of `PrimInt`, so adding new constraints would break others, however theoretical such others may be.

src/ops/checked.rs
```diff
@@ -90,3 +90,54 @@ checked_impl!(CheckedDiv, checked_div, i32);
 checked_impl!(CheckedDiv, checked_div, i64);
 checked_impl!(CheckedDiv, checked_div, isize);
 
+/// Performs a left shift that returns `None` on overflow.
+pub trait CheckedShl<RHS>: Sized + Shl<RHS, Output=Self> {
```
cuviper: I think for now, I'd rather keep these consistent with the other `Checked` traits that are not parameterized, especially since the standard library methods are only for `u32`.  That also moots your comment below about needing checked conversions to `u32`.

(We could decide to expand it later using a default, like `CheckedShl<RHS = u32>`.)

fabianschuiki:
How about now?

cuviper:
Looks good, thanks!

bors r+

PR #22, opened 2018-01-04
Implement std::fmt::Display for ParseFloatError

### LINKED ISSUE #2: std::fmt::Display is not implemented for num::traits::ParseFloatError
_From @spease on September 2, 2017 19:24_

ParseFloatError in the stdlib provides a Display implementation. num implements its own ParseFloatError, but does not provide a Display implementation.

_Copied from original issue: rust-num/num#332_

svartalf:
Basically it is a copy of the stdlib implementation, should close #2


cuviper:
Thanks!

bors r+

PR #24, opened 2018-01-12
"std" feature

termoshtt:
Copy of "next" branch in num
A clone of the original PR https://github.com/rust-num/num/pull/296 with some fixes in `ci/`
Ref #16

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,3 +12,7 @@ version = "0.1.41"
 readme = "README.md"
 
 [dependencies]
+
+[features]
+default = ["std"]
+std = []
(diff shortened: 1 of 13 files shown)


cuviper:
I appreciate your initiative, but I was hoping to do this in a way that preserves some authorship from before.  I also don't want to include breaking changes like `One::is_one` yet, and I'm reconsidering my former reviews in rust-num/num#296 where I suggested `#[cfg]` on methods and said:

> We should probably also add a caution in the docs for implementors of Float. Those that implement it normally are fine, but if someone now chooses to implement Float only for no_std, their build will fail if anyone else in the entire dependency graph enables num-traits/std.

In retrospect, this seems like something that will inevitably blow up.  Maybe their original `BaseFloat` idea was enough, or some variation like the `Real` in #23.  I'm tempted to just make `Float` depend on `std` with no baser alternative at all for now, and we can add something back in later.

termoshtt:
> but I was hoping to do this in a way that preserves some authorship from before

Your are quite right. I close this PR, and read the origitnal PR more carefully...

PR #25, opened 2018-01-17
Replaced fn one() and fn zero() by associated consts

porky11:
The new consts, every trait One and Zero have to implement are called ONE and ZERO.
For backwards compatability, the functions are not deleted, so using such types will still be possible as before.
But implementors, who update the version would need to update their trait implementations, using these consants.
It's not possible to define the functions as associated const fn, and set them as default value for the new constants.
So I updated the version to `0.2.0`.
I also removed the fixme messages for #5527
In order to reduce breaking changes, I also updated the bounded trait and it's impls
New implementations will only have to implement the constants, the functions have default implementations retruning the constant.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ categories = [ "algorithms", "science" ]
 license = "MIT/Apache-2.0"
 repository = "https://github.com/rust-num/num-traits"
 name = "num-traits"
-version = "0.1.41"
+version = "0.2.0"
 readme = "README.md"
 
 [dependencies]
--- a/src/bounds.rs
+++ b/src/bounds.rs
@@ -5,21 +5,25 @@ use std::num::Wrapping;
 
 /// Numbers which have upper and lower bounds
 pub trait Bounded {
-    // FIXME (#5527): These should be associated constants
     /// returns the smallest finite number this type can represent
-    fn min_value() -> Self;
+    const MIN: Self;
+    #[inline]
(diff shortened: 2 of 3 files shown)


cuviper:
Is it even _possible_ for heap-allocated types like `BigUint` to define a `const`?  I think not, which makes this change a non-starter.  I think we should just remove the "fixme" comments you found.

And more generally speaking, I fear that such breaking changes to num-traits are going to be too much pain to be worth it.  If we ever do break, it should be a very rare event, and there are probably a lot of things that ought to be redesigned in the process.

porky11:
If heap-allocated types cannot be const, this would be bad.
But associated consts seem like a better solution.
If breaking changes already are pain, why is this crate not version 1.x.y?

porky11:
I'd use associated consts most frequently with number types, and when wanting to be generic, the associated constants would become useless without a generic one or zero

cuviper:
> If heap-allocated types cannot be const, this would be bad.
> But associated consts seem like a better solution.

It's a trade-off.  A true constant can be used in more ways, but can't be implemented by non-const types.

> If breaking changes already are pain, why is this crate not version 1.x.y?

History and momentum.  This crate is widely used as a public dependency (meaning that it's exposed in the public API of other crates), so I've chosen to maintain it as a stable crate, despite the 0.1.x version.  The semantics of a version are only what you make of it.

porky11:
Maybe const fns and constants will allow more things some time, then it may be worth thinking about it another time.

Issue #12, opened 2017-12-19
Casting floats to integers does not check overflow


cuviper:
_From @photino on October 4, 2015 15:49_

The following code 

``` rust
num::cast::<f64, i32>(1.0e+123f64)
```

will return

``` rust
Some(-2147483648)
```

instead of `None`.

_Copied from original issue: rust-num/num#119_

cuviper:
I think the root problem is `impl_to_primitive_float`, which just typecasts to integers with `as`.

cuviper:
_From @apopiak on November 25, 2015 18:49_

I'd like to try this one.

cuviper:
@apopiak go for it!

cuviper:
Technically this is UB -- rust-lang/rust#10184 -- and I believe rustc is planning to implement saturating float casts.  I think even with that, our conversion traits should prefer `None` instead of saturating.

apopiak:
Yeah, sorry I never got back on that. Consider me no longer working on this until I post otherwise.

dbarella:
I think I have a patch that fixes this!

PR #28, opened 2018-01-27
Check overflow when casting floats to integers.

dbarella:
This change adds some new macro rules used when converting from floats
to integers. There are two macro rule variants, one for signed ints, one
for unsigned ints.

Among other things, this change specifically addresses the overflow case
documented in https://github.com/rust-num/num-traits/issues/12

### DIFF
--- a/src/cast.rs
+++ b/src/cast.rs
@@ -655,9 +655,4 @@ fn float_to_integer_checks_overflow() {
 
     // Expect the overflow to be caught
     assert_eq!(source.to_i32(), None);
-
-    // Specifically make sure we didn't silently let the overflow through
-    // (This line is mostly for humans -- the robots should catch any problem
-    // on the line above).
-    assert!(source.to_i32() != Some(-2147483648));
 }
--- a/src/cast.rs
+++ b/src/cast.rs
@@ -236,35 +236,93 @@ macro_rules! impl_to_primitive_float_to_float {
     )
 }
 
+macro_rules! impl_to_primitive_float_to_signed_int {
+    ($SrcT:ident, $DstT:ident, $slf:expr) => (
+        if size_of::<$SrcT>() <= size_of::<$DstT>() {
+            Some($slf as $DstT)
+        } else {
+            // Make sure the value is in range for the cast.
+            let n = $slf as $DstT;
+            let max_value: $DstT = ::std::$DstT::MAX;
+            if -max_value as $DstT <= n && n <= max_value as $DstT {
+                Some($slf as $DstT)
+            } else {
+                None
+            }
+        }
+    )
+}
+
+macro_rules! impl_to_primitive_float_to_unsigned_int {
+    ($SrcT:ident, $DstT:ident, $slf:expr) => (
+        if size_of::<$SrcT>() <= size_of::<$DstT>() {
+            Some($slf as $DstT)
+        } else {
+            // Make sure the value is in range for the cast.
+            let n = $slf as $DstT;
+            let max_value: $DstT = ::std::$DstT::MAX;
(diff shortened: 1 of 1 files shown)


dbarella:
Ah, I see that CI is spinning against an version of rust which doesn't provide `assert_ne!` -- I'll change that.

dbarella:
As a general note, I'm currently a Google employee, which means that Google owns the copyright for this patch by default. There is a process internal to Google by which employees can ask Google to liberate the copyright -- I intend to start that process first thing on Monday. But I wanted to leave a comment here to let y'all know.

*I will not merge this PR without an explicit green light from y'all* -- I just want to be as upfront as possible about what's going on behind the scenes ^_^

cuviper:
RE: Google, we don't ask for copyright assignment, so as long as your contribution is compatible with the project licenses (MIT and Apache-2.0), I think we're fine. IANAL, of course -- please do confirm with Google if you're not sure about this yourself.

src/cast.rs
```diff
@@ -589,3 +647,17 @@ fn as_primitive() {
...
+    // This will overflow an i32
+    let source: f64 = 1.0e+123f64;
+
+    // Expect the overflow to be caught
+    assert_eq!(source.to_i32(), None);
+
+    // Specifically make sure we didn't silently let the overflow through
+    // (This line is mostly for humans -- the robots should catch any problem
+    // on the line above).
+    assert!(source.to_i32() != Some(-2147483648));
```
cuviper: As a human, I don't understand the point of this line.  If it was equal to `None` above, then it won't be equal to _any_ `Some`.  Trust the "robots".

Also note that -2147483648 is exactly `i32::MIN`, which would be a perfectly legal value!  It just happens that LLVM is giving you this value for overflow too, but that's UB.

dbarella:
Overall: Understood! It seems that the direction of both of the PRs is pretty similar, though they obviously differ in the core macro definition which checks for overflow. I'd love to cherry-pick the commits from @apopiak's PR, but I doubt I'll be able to apply them cleanly -- they were written to `traits.rs`, not `cast.rs`, and after a half-hour of hacking on different approaches, I haven't gotten anywhere. Any recommendations on how to do that?

src/cast.rs
dbarella: I figured it might be nice to have some visual correspondence between this test and the PR, but I agree that it's "nonsensical" otherwise -- we shouldn't ever get to this line. I've removed it.

cuviper:
Yes, `commit --author` is fine if cherry-pick isn't working.  FWIW, you'll be recorded as Committer either way.

dbarella:
Indeed, authorship attribution is the main thing I'm going for. I'll do that then.

cuviper:
BTW, you might want to keep an eye on rust-lang/rust#47857.  The current implementation isn't sound, and they're debating whether `TryFrom` should do float to integer _at all_, but if they get past that then it may result in something we can imitate too.  Maybe you'd even like to get involved there?

dbarella:
Affirmative, thanks for the heads up! I'll comment there and see what I can contribute.

I spent a little more time today hacking on this PR, nothing major. It's non-functional right now, so don't worry about reviewing anything new yet. Among other things, I added apopiak@'s changes in using `--author=...`, so the attribution is there now. I'm also starting to look at the boundary conditions that you mentioned before -- currently working on some tests to make sure that we check all the right things. Definitely open to any suggestions you might have! I'm hoping to check back in a couple days with a more functional implementation.

cuviper:
I've continued this in #52 -- can you take a look?

cuviper:
I merged #52, but further code reviews are welcome at any time!

PR #30, opened 2018-02-01
Re-introduce the std feature

### LINKED ISSUE #16: Make std a feature, using `#![no_std]` otherwise

cuviper:
This is a port of @vks's rust-num/num#296, but without the feature-toggled changes to `Float`.  Now `Float` and the newer `Real` are completely dependent on having `std` enabled.  In the future we can consider adding separate more-limited float/real traits that can work without `std`, like the `BaseFloat` that was originally proposed in the former PR.

This is a breaking change with a bump to 0.2, since anyone currently using `default-features = false` will lose functionality.  The actual API is otherwise unchanged, so my plan is to employ the "semver trick" -- publishing a new num-traits-0.1 that re-exports everything from 0.2 (with `std`).  Thus all `num-traits` users should remain compatible even if they mix 0.1 and 0.2.

Closes #16.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ categories = [ "algorithms", "science" ]
 license = "MIT/Apache-2.0"
 repository = "https://github.com/rust-num/num-traits"
 name = "num-traits"
-version = "0.1.42"
+version = "0.2.0-pre"
 readme = "README.md"
 
 [dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,3 +12,7 @@ version = "0.1.42"
 readme = "README.md"
 
 [dependencies]
+
+[features]
+default = ["std"]
+std = []
--- a/README.md
+++ b/README.md
@@ -21,6 +21,19 @@ and this to your crate root:
 extern crate num_traits;
 ```
 
+## Features
+
+This crate can be used without the standard library (`#![no_std]`) by disabling
+the default `std` feature.  Use this in `Cargo.toml`:
+
+```toml
+[dependencies.num-traits]
+version = "0.1"
+default-features = false
+```
+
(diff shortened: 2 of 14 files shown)


cuviper:
cc @termoshtt, since you had opened #24 before.

cuviper:
@vks, I wouldn't call it a regression.  The previous `no_std` PR was revealed to be a breaking change, so it was yanked and reverted.  Thus right now, there isn't any `no_std` num-traits.

But yes, I've come around that `BasicFloat` was probably the right way to expose *some* functionality after all.  I'm open to adding that again, but it doesn't have to be part of this initial `no_std` change.  Would you like to send that anew after I merge this one?

vks:
> But yes, I've come around that BasicFloat was probably the right way to expose some functionality after all. I'm open to adding that again, but it doesn't have to be part of this initial no_std change. Would you like to send that anew after I merge this one?

With pleasure, but it will be a breaking change if we introduce `Float: BasicFloat`. (Maybe `CoreFloat` is actually a better name.)

cuviper:
> it will be a breaking change if we introduce `Float: BasicFloat`.

Yes, it needs to be independent, even if we did it right away here.  I'm not making any API changes, so I can compatibly re-export everything back to 0.1.  The *only* breaking change is the lost functionality under `no_std`, which is something one can choose when updating to 0.2.

The `CoreFloat` name does sound good, as it gives some hint about why this should exist at all.

vks:
Alright, then everything looks good to me!

bors r+

cuviper:
bors r=vks

PR #31, opened 2018-02-01
Re-export all items from num-traits 0.2

cuviper:


Cargo.toml
```diff
@@ -8,7 +8,12 @@ categories = [ "algorithms", "science" ]
...
+version = "0.1.43"
 readme = "README.md"
 
-[dependencies]
+[lib]
+doctest = false # multiple rlib candidates for `num_traits` found
+
+[dependencies.num-traits]
+git = "https://github.com/cuviper/num-traits"
+branch = "no_std"
```
cuviper: This is the git and branch for #30.  After that's merged and published, this dependency should become `version = "0.2"`.

PR #32, opened 2018-02-02
Implement CoreFloat trait

vks:
This is a subset of the `Float` trait, but works with `no_std`.
Some code was simplified by using `CoreFloat`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,7 +26,7 @@ use core::fmt;
 pub use bounds::Bounded;
 #[cfg(feature = "std")]
 pub use float::Float;
-pub use float::FloatConst;
+pub use float::{CoreFloat, FloatConst};
 // pub use real::Real; // NOTE: Don't do this, it breaks `use num_traits::*;`.
 pub use identities::{Zero, One, zero, one};
 pub use ops::checked::{CheckedAdd, CheckedSub, CheckedMul, CheckedDiv, CheckedShl, CheckedShr};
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -26,7 +26,7 @@ use core::fmt;
 pub use bounds::Bounded;
 #[cfg(feature = "std")]
 pub use float::Float;
-pub use float::FloatConst;
+pub use float::{CoreFloat, FloatConst};
 // pub use real::Real; // NOTE: Don't do this, it breaks `use num_traits::*;`.
```
cuviper: I think we may have the same concern as `Real` did here.  If someone does `use num_traits::*`, then the common methods between `Float` and `CoreFloat` may be ambiguous.  Would you mind if we didn't re-export this in the root?

vks: I'm not sure supporting wildcard imports is worthwhile. If you want to make sure not to break them, you can't add any new functions or types. However, in this case I guess it is easy enough to avoid that problem.

cuviper: I'm not sure either.  I guess a wildcard making us self-conflicting is a worse than having a conflict with some unrelated crate though.

vks:
I think I addressed all your comments.

cuviper:
Looks great, thanks!

bors r+

cuviper:
@vks I just noticed that this doesn't require `NumCast` like `Float` does.  Was that intentional?

vks:
Yes, I did not need it. Is there a reason to require it?

cuviper:
It's a benefit for those who use the trait too. It was there long before
Float itself needed it.

I think I'm going to add a few more methods from Float that are feasible
with no_std, before it's published and becomes a breaking change. I can add
NumCast as I do that...

PR #37, opened 2018-02-18
Add Inv and Pow traits.

### LINKED ISSUE #34: Feature request: Recip trait
Essentially, a pair to `Neg` for multiplication. Anything which can have its reciprocal taken would implement this. Useful in the context of `Mul + Recip + One`, for example.

clarcharr:
This is not a breaking change, and closes #34 and #38.

This doesn't add any impls for the other `num` crates, just floats with `std` enabled. The trait has to be added before those other crates can be updated.


src/ops/mod.rs
```diff
@@ -1,3 +1,4 @@
 pub mod saturating;
 pub mod checked;
 pub mod wrapping;
+pub mod new;
```
cuviper: Should be `pub mod inv;`?

src/ops/inv.rs
```diff
@@ -0,0 +1,29 @@
...
+}
+
+macro_rules! inv_impl {
+    ($t:ty, $out:ty, $fn:expr) => {
+        impl<'a> Inv for $t {
+            type Output = $out;
+
+            #[inline]
+            fn inv(self) -> $out {
+                ($fn)(self)
```
cuviper: I don't this function call will actually work for the reference versions, and it's weird having a lifetime parameter on the value versions.  There are only 4 implementations here, so maybe it's not worth trying to make a unified macro.

To avoid `cfg(feature = "std")`, we could just implement these as `1.0 / self`.  That's all the standard library is doing anyway.

clarcharr: I'll do that later today

src/ops/mod.rs
clarcharr: Yes

clarcharr:
I figured I'd implement #38 in this as well considering how I was going to offer both, and I don't want to deal with merge conflicts. Let me know if everything looks alright!

src/ops/pow.rs
```diff
@@ -0,0 +1,71 @@
...
+pow_impl!(i16, u32, pow);
+pow_impl!(u32, u32, pow);
+pow_impl!(i32, u32, pow);
+pow_impl!(u64, u32, pow);
+pow_impl!(i64, u32, pow);
+pow_impl!(usize, u32, pow);
+pow_impl!(isize, u32, pow);
+
+#[cfg(feature = "std")]
+mod float_impls {
```
cuviper: You'll have to `use super::Pow;` to get it into this module.

cuviper:
Let's put `Pow` into the existing `pow` module, instead of under `ops`.

src/ops/pow.rs
clarcharr: …that is correct. I'll just factor out the `cfg`.

src/ops/inv.rs
```diff
@@ -0,0 +1,40 @@
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use num_traits::{Inv, One};
+    ///
+    /// let x = 7.0;
+    /// let y = -0.0;
+    /// assert_eq!(x.inv() * x, One::one());
+    /// assert_eq!(y.inv() * y, One::one());
```
cuviper: This is `-inf * -0.0`, which is actually `NAN`.  Maybe just assert the `-inf` instead?

src/pow.rs
```diff
@@ -1,6 +1,81 @@
...
+    /// The result after applying the operator.
+    type Output;
+
+    /// Returns `self` to the power `rhs`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use num_traits::Pow;
+    /// assert_eq!(10.pow(2), 100);
```
cuviper: CI failed here because it inferred this like `10i32.pow(2i32)`, which is not implemented.

clarcharr:
Should be ready to merge now.

cuviper:
Sorry if I misspoke, but I did not mean that we should have so many variations of `Pow` implementations.  Do you have your own reasons for all them?  I'm especially worried about the ones with lossy `as` casts.  Just the basic natively-supported `pow` implementations looked good to me.

clarcharr:
@cuviper I accidentally added the `u64` casts in error, but the rest are lossless casts (such as `u8 as u32`).

My main reasoning was to do something similar to how `Shl` is done in the standard library: https://doc.rust-lang.org/std/ops/trait.Shl.html

cuviper:
Ah, I got confused by the macros, and thought you had some signed->unsigned casts in there too.

Can we do it without the closures though?  I think if you leverage `From`/`Into`, then the final implementation can be just `$method(self, rhs.into())`, where `$method` is either the type's inherent `pow` or num's `pow` for `usize`.  No `as` casting at all.

Should floats do `u32` and `usize` with num's `pow` too?  And perhaps also `u8` and `u16` this way under `no_std` (without `powi`)?

I count 372 `Pow` implementations so far in the generated docs!  The downside is that all this does significantly increase the compilation time, compared to the current `master` branch.

clarcharr:
I think that doing without the closures is very fair and I'll refactor this in a bit to achieve that. Also I just noticed rust-lang/rust#48321 and I think that simply deferring to `wrapping_pow` once it's stable would be better for `Wrapping`, and that those impls can just be removed for now to avoid compile time hikes.

I'll see what I can do about benchmarking compile times and how that can be improved.

cuviper:
To put real numbers on it, `cargo build --release` goes from 0.72 seconds to 1.3 seconds on my i7-7700k.  That's still bearable, but quite a significant increase for just one feature.

clarcharr:
I don't think a double in compile time is acceptable; I'll see what I can do.

clarcharr:
@cuviper with the latest change, my compile time went from 2.29s to 1.84s, which is only a 20% increase. Could you provide benchmarks on your end?

Longer-term it may be best to make the standalone `pow` function private and change the `usize` exponent to a `u64` for compatibility. This is breaking, obviously, and I didn't do that.

cuviper:
For me it's now 0.86s vs master's 0.72s -- that seems fine.  There are still a lot of implementations -- 292 -- so I'd guess the closures were the main culprit for the slowdown.

Let's merge!

bors r+

clarcharr:
Awesome! Could you push a new version of `num-traits` with these changes added?

cuviper:
Yeah, but I still want to revisit `FloatCore` and perhaps add some more to it before it's published.  I'll try to do this today.

clarcharr:
That's fair; no rush. :)

Issue #5, opened 2017-12-19
Add method `is_one` to `One` trait


cuviper:
_From @pyfisch on August 5, 2016 17:15_

The [`Zero` trait](http://rust-num.github.io/num/num/trait.Zero.html) contains two methods: `zero` returns the number 0 for the given type and `is_zero` is provided and can be used to check if a number is 0.

The [`One` trait](http://rust-num.github.io/num/num/trait.One.html) is similar but contains only the method `one` to create the number 1. Please also add a method to check if a number is one.

_Copied from original issue: rust-num/num#214_

cuviper:
It turned out not so easy without a breaking change, although we could add it `where Self: PartialEq` for now to allow a default implementation, and drop that in a breaking change (eventually).

cuviper:
That was rust-num/num#336 on the `next` branch.

clarcharr:
The issue description/title should be updated to mention removing the `PartialEq` bound rather than adding the function.

vks:
Shouldn't this be closed?

cuviper:
I added the `PartialEq` note to #47 for breaking changes.

PR #39, opened 2018-02-19
Add is_one.

clarcharr:
Implements the version recommended in #5. That issue should remain open to track the breaking-change version.

### DIFF
--- a/src/identities.rs
+++ b/src/identities.rs
@@ -79,6 +79,14 @@ pub trait One: Sized + Mul<Self, Output = Self> {
     /// `static mut`s.
     // FIXME (#5527): This should be an associated constant
     fn one() -> Self;
+
+    /// Returns `true` if `self` is equal to the multiplicative identity.
+    ///
+    /// Compatibility note: this method will be a requirement to implement in the future; new
+    /// implementors should provide it for future-compatibility.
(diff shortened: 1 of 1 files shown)


src/identities.rs
```diff
@@ -79,6 +79,14 @@ pub trait One: Sized + Mul<Self, Output = Self> {
     /// `static mut`s.
     // FIXME (#5527): This should be an associated constant
     fn one() -> Self;
+
+    /// Returns `true` if `self` is equal to the multiplicative identity.
+    ///
+    /// Compatibility note: this method will be a requirement to implement in the future; new
+    /// implementors should provide it for future-compatibility.
```
cuviper: I'm a little hesitant about this scary-sounding note, because we're definitely *not* going to change that without a semver bump!  And when we do, it should also drop the `Self: PartialEq`, which means they'll have porting work to do anyway.  Maybe instead just advise that implementations should consider adding it now for better performance, noting that it will only become required in a future breaking change.

Also, let's put `#[inline]` on this.

clarcharr:
Applied your suggestions; is the caveat better now?

cuviper:
Works for me, thanks!

bors r+

PR #41, opened 2018-02-28
Various improvements to FloatCore

cuviper:
- New macros simplify forwarding method implementations.
  - `Float` and `Real` use this to compact their implementations.
  - `FloatCore` now forwards `std` implementations when possible.
- `FloatCore` now requires `NumCast`, like `Float does.
- New additions to `FloatCore`:
  - Constants like `min_value()` -> `f64::MIN`
  - Rounding methods `floor`, `ceil`, `round`, `trunc`, `fract`
  - `integer_decode` matching `Float`'s
- Fix NAN sign handling in `FloatCore` (rust-num/num#312, rust-lang/rust#42425)
- Fix overflow in `FloatCore::powi` exponent negation.
- Add doctests to all `FloatCore` methods.

### DIFF
--- a/src/float.rs
+++ b/src/float.rs
@@ -5,7 +5,7 @@ use core::num::FpCategory;
 use core::f32;
 use core::f64;
 
-use {Num, NumCast};
+use {Num, NumCast, ToPrimitive};
 
 /// Generic trait for floating point numbers that works with `no_std`.
 ///
@@ -668,7 +668,9 @@ pub trait FloatCore: Num + NumCast + Neg<Output = Self> + PartialOrd + Copy {
             self = self.recip();
         }
         // It should always be possible to convert a positive `i32` to a `usize`.
-        super::pow(self, exp as u32 as usize)
+        // Note, `i32::MIN` will wrap and still be negative, so we need to convert
+        // to `u32` without sign-extension before growing to `usize`.
+        super::pow(self, (exp as u32).to_usize().unwrap())
     }
 
     /// Converts to degrees, assuming the number is in radians.
--- a/src/float.rs
+++ b/src/float.rs
@@ -65,6 +65,76 @@ pub trait FloatCore: Num + NumCast + Neg<Output = Self> + PartialOrd + Copy {
     /// predicate instead.
     fn classify(self) -> FpCategory;
 
+    /// Returns the largest integer less than or equal to a number.
+    #[inline]
+    fn floor(self) -> Self {
+        let f = self.fract();
+        if f.is_nan() || f.is_zero() {
+            self
+        } else if self < Self::zero() {
+            self - f - Self::one()
+        } else {
+            self - f
+        }
+    }
+
+    /// Returns the smallest integer greater than or equal to a number.
+    #[inline]
+    fn ceil(self) -> Self {
+        let f = self.fract();
+        if f.is_nan() || f.is_zero() {
+            self
+        } else if self > Self::zero() {
+            self - f + Self::one()
+        } else {
+            self - f
+        }
+    }
+
+    /// Returns the nearest integer to a number. Round half-way cases away from `0.0`.
+    #[inline]
+    fn round(self) -> Self {
+        let one = Self::one();
+        let h = Self::from(0.5).expect("Unable to cast from 0.5");
+        let f = self.fract();
+        if f.is_nan() || f.is_zero() {
+            self
+        } else if self > Self::zero() {
+            if f < h {
+                self - f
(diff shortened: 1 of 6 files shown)


cuviper:
cc #32 @vks - I'd really appreciate a review on this, if only to make sure I didn't butcher your work. 😅

src/float.rs
```diff
@@ -1493,8 +1493,8 @@ mod tests {
             assert!((FloatCore::to_radians(deg) - rad).abs() < 1e-6);
 
             let (deg, rad) = (deg as f32, rad as f32);
-            assert!((FloatCore::to_degrees(rad) - deg).abs() < 1e-6);
-            assert!((FloatCore::to_radians(deg) - rad).abs() < 1e-6);
+            assert!((FloatCore::to_degrees(rad) - deg).abs() < 1e-5);
+            assert!((FloatCore::to_radians(deg) - rad).abs() < 1e-5);
```
vks: Why was this necessary?

src/float.rs
```diff
@@ -164,14 +505,54 @@ pub trait FloatCore: Num + NumCast + Neg<Output = Self> + PartialOrd + Copy {
     }
 
     /// Returns `true` if `self` is positive, including `+0.0` and
-    /// `FloatCore::infinity()`.
+    /// `FloatCore::infinity()`, and with newer versions of Rust
+    /// even `FloatCore::nan()`.
```
vks: Can we be explicit about the Rust version?

cuviper: Yeah, I'll look that up.

src/float.rs
```diff
@@ -218,11 +218,11 @@ pub trait FloatCore: Num + NumCast + Neg<Output = Self> + PartialOrd + Copy {
     #[inline]
     fn powi(mut self, mut exp: i32) -> Self {
         if exp < 0 {
-            exp = -exp;
+            exp = exp.wrapping_neg();
             self = self.recip();
         }
         // It should always be possible to convert a positive `i32` to a `usize`.
-        super::pow(self, exp.to_usize().unwrap())
+        super::pow(self, exp as u32 as usize)
```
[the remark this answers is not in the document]
cuviper: To handle `i32::MIN`. See above that I switched it to `wrapping_neg`, which will leave `MIN` the same. We want that to become an unsigned positive 0x80000000, but `to_usize()` would just fail.

src/float.rs
cuviper: After I made it forward to std's `f32::to_degrees`, it started failing on nightly. I suspect rust-lang/rust#47919, which tried to be more accurate, actually made things *less* accurate for these inputs, but I haven't investigated yet. I will probably file an upstream bug.

cuviper:
> What is the motivation behind adding NumCast to FloatCore? Just symmetry with Float? I think the traits are kind of orthogonal. It seems like you need NumCast for the implementation of round.

For symmetry, yes.  I thought it made sense to keep `Float` and `FloatCore` as similar as possible.  Do you see any situation where it's a *problem* to have it?

I did end up using `NumCast` for `round`, but just for a simple 0.5 -- we could avoid that with `one() / (one() + one())` if necessary.  Or we could move that implementation of `round` directly to `f32` and `f64` (`no_std` only), leaving no default.

> Should we use associated constants rather than static methods? Would that raise our minimal supported Rust version?

It would raise it a lot -- associated consts were only made stable in 1.20.

> Are the implementations of the rounding functions canonical? I don't know enough about the properties of floats to vouch for their correctness. It seems like tests for them are missing?

I implemented them independently based on their description and observed behavior.  I'm fairly confident in them, but not 100% sure.  They do have doctests where I tried to exercise interesting cases, and note this gets tested both with and without forwarding to `std`, so I know we're at least consistent for the inputs I selected.

src/float.rs
cuviper: https://github.com/rust-lang/rust/issues/48617

src/float.rs
cuviper: Changed the comment to mention that this was Rust 1.20.

src/float.rs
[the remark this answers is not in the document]
cuviper: I compromised -- added a comment why the `as u32` is needed, but went back to using `to_usize()` so we won't be surprised here by a platform with 16-bit `usize` (although that will panic).

cuviper:
I'm going to merge this so we can publish.  The rounding parts that I'm less confident about can always be bug-fixed later.  Thanks for reviewing!

bors r=vks

PR #43, opened 2018-03-04
Add Default constraint to PrimInt

jturner314:
I don't see any reason to intentionally leave off a `Default` trait bound, and it's useful to have. Note that this is technically a breaking change for anyone who has implemented `PrimInt` for their own types, so it should probably be part of `num-traits` 0.3 instead of 0.2.2.

### DIFF
--- a/src/int.rs
+++ b/src/int.rs
@@ -8,6 +8,7 @@ use ops::saturating::Saturating;
 pub trait PrimInt
     : Sized
     + Copy
+    + Default
     + Num + NumCast
     + Bounded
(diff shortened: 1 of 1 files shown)


cuviper:
I have no short-term plans to make breaking changes.  Rather than leaving a PR in limbo, can you file a tracking issue for this?  #36 also wanted `Display` and `Debug`.

jturner314:
Okay. I created issue #44.

cuviper:
Thanks!

PR #59, opened 2018-04-09
Added `MulAdd` and `MulAddAssign` traits

regexident:
Both `f32` and `f64` implement fused multiply-add, which computes `(self * a) + b` with only one rounding error. This produces a more accurate result with better performance than a separate multiplication operation followed by an add:

```rust
fn mul_add(self, a: f32, b: f32) -> f32[src]
```

It is however not possible to make use of this in a generic context by abstracting over a trait.

My concrete use-case is machine learning, [gradient descent](https://en.wikipedia.org/wiki/Gradient_descent) to be specific,  
where the core operation of updating the gradient could make use of `mul_add` for both its `weights: Vector` as well as its `bias: f32`:

```rust
struct Perceptron {
  weights: Vector,
  bias: f32,
}

impl MulAdd<f32, Self> for Vector {
  // ...
}

impl Perceptron {
  fn learn(&mut self, example: Vector, expected: f32, learning_rate: f32) {
    let alpha = self.error(example, expected, learning_rate);
    self.weights = example.mul_add(alpha, self.weights);
    self.bias = self.bias.mul_add(alpha, self.bias)
  }
}
```

(The actual impl of `Vector` would be generic over its value type: `Vector<T>`, thus requiring the trait.)

### DIFF
--- a/src/ops/mul_add.rs
+++ b/src/ops/mul_add.rs
@@ -111,7 +111,7 @@ mod tests {
                         let x: $t = 3;
                         let b: $t = 4;
 
-                        assert_eq!(m.mul_add(x, b), (m*x + b));
+                        assert_eq!(MulAdd::mul_add(m, x, b), (m*x + b));
                     }
                 )+
             };
@@ -121,6 +121,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(feature = "std")]
     fn mul_add_float() {
         macro_rules! test_mul_add {
             ($($t:ident)+) => {
@@ -132,7 +133,7 @@ mod tests {
                         let x: $t = 3.4;
                         let b: $t = 5.6;
 
-                        let abs_difference = (m.mul_add(x, b) - (m*x + b)).abs();
+                        let abs_difference = (MulAdd::mul_add(m, x, b) - (m*x + b)).abs();
 
                         assert!(abs_difference <= $t::EPSILON);
                     }
--- a/src/ops/mul_add.rs
+++ b/src/ops/mul_add.rs
@@ -1,0 +1,157 @@
+/// The fused multiply-add operation.
+/// Computes (self * a) + b with only one rounding error.
+/// This produces a more accurate result with better performance
+/// than a separate multiplication operation followed by an add.
+///
+/// Note that `A` and `B` are `Self` by default, but this is not mandatory.
+///
+/// # Example
+///
+/// ```
+/// use std::f32;
+///
+/// let m = 10.0_f32;
+/// let x = 4.0_f32;
+/// let b = 60.0_f32;
+///
+/// // 100.0
+/// let abs_difference = (m.mul_add(x, b) - (m*x + b)).abs();
+///
+/// assert!(abs_difference <= f32::EPSILON);
+/// ```
+pub trait MulAdd<A = Self, B = Self> {
+    /// The resulting type after applying the fused multiply-add.
+    type Output;
+
+    /// Performs the fused multiply-add operation.
+    fn mul_add(self, a: A, b: B) -> Self::Output;
+}
+
+/// The fused multiply-add assignment operation.
+pub trait MulAddAssign<A = Self, B = Self> {
+    /// Performs the fused multiply-add operation.
+    fn mul_add_assign(&mut self, a: A, b: B);
+}
+
+impl MulAdd<f32, f32> for f32 {
+    type Output = Self;
+
+    #[inline]
+    fn mul_add(self, a: Self, b: Self) -> Self::Output {
+        if cfg!(feature = "std") {
+            f32::mul_add(self, a, b)
+        } else {
+            (self * a) + b
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


cuviper:
Is this actually useful to you for integers?  Otherwise, you can just use [`Float::mul_add`](https://docs.rs/num-traits/0.2.2/num_traits/float/trait.Float.html#tymethod.mul_add).

src/ops/mul_add.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+
+impl MulAdd<f64, f64> for f64 {
+    type Output = Self;
+
+    #[inline]
+    fn mul_add(self, a: Self, b: Self) -> Self::Output {
+        #![allow(unconditional_recursion)]
+        f64::mul_add(self, a, b)
```
cuviper: I believe this `allow` is masking a real problem.  In a `#![no_std]` build, `f64` doesn't have most of its inherent methods, which means this call will resolve back to `MulAdd::mul_add` again -- thus the unconditional recursion.

regexident:
> Is this actually useful to you for integers? Otherwise, you can just use `Float::mul_add`.

It's useful for writing code that makes use of the optimization if applicable, yet remains fully  generic. A particular use case is fixed-point arithmetic on `T: Real + MulAdd` as a convenient drop-in replacement for `T: Float + MulAdd`.

src/ops/mul_add.rs
regexident: Oh `#![no_std]` is what made the builds fail on CI, yet succeed on my local machine.
Adding the `#![allow(unconditional_recursion)]` felt fishy.

How would you propose solving this? Something like this?

```rust
if cfg!(feature == "std") {
    f64::mul_add(self, a, b)
} else {
    (self * a) + b
}
```

vks:
> his produces a more accurate result with better performance than a separate multiplication operation followed by an add

Unfortunately, in my experience this is not necessarily true. For `f64` I don't see a performance difference on my machine, and for `f32` it is slower without `target-cpu=native` and faster with.

regexident:
> Unfortunately, in my experience this is not necessarily true.

This does not invalidate the desire to have a way to generically express `fn mul_add` through a trait though, or does it?

vks:
No, performance is not really related to the changes in this PR. What might be problematic is that `a * b + c` is less precise than `a.mul_add(b, c)` and will yield different results. I'm not sure how big of a problem that is in practice, but it should probably be documented.

Also see https://github.com/rust-lang/rust/pull/44805.

src/ops/mul_add.rs
cuviper: I think it would be better to *not* implement the trait at all for floats in `#![no_std]` mode, so we don't mislead about the rounding errors.  Either that, or do a manual implementation with appropriate rounding fixups, but I imagine that's non-trivial.

regexident:
> No, performance is not really related to the changes in this PR. What might be problematic is that `a * b + c` is less precise than `a.mul_add(b, c)` and will yield different results. I'm not sure how big of a problem that is in practice, but it should probably be documented.

Good point, I'll gladly add a mention of this to the documentation. :)

Any further feedback? What needs to be done to proceed? Do we want to proceed?

cuviper:
> > Is this actually useful to you for integers? Otherwise, you can just use `Float::mul_add`.
> 
> It's useful for writing code that makes use of the optimization if applicable, yet remains fully generic. A particular use case is fixed-point arithmetic on `T: Real + MulAdd` as a convenient drop-in replacement for `T: Float + MulAdd`.

Note there's [`Real::mul_add`](https://docs.rs/num-traits/0.2.2/num_traits/real/trait.Real.html#tymethod.mul_add) too.

Supporting the `no_std` fpa crate is a more compelling example though, since `Float` and `Real` are only available in `std` builds.

I still think we should *not* have a `(self * a) + b` fallback for `no_std` floats -- I'd rather not implement the trait for `no_std` floats at all if we can't meet the same rounding accuracy.  The performance is secondary, maybe not even worth bringing up.

regexident:
I just moved the impls of `MulAdd`/`MulAddAssign` for `f32`/`f64` behind a feature guard with commit 28be885.

src/ops/mul_add.rs
```diff
@@ -0,0 +1,145 @@
...
+            ($($t:ident)+) => {
+                $(
+                    {
+                        use core::$t;
+
+                        let m: $t = 12.0;
+                        let x: $t = 3.4;
+                        let b: $t = 5.6;
+
+                        let abs_difference = (m.mul_add(x, b) - (m*x + b)).abs();
```
cuviper: I'm surprised this test didn't fail on `no_std`, but I think it's because `libtest` links to `libstd`, so the compiler still has the inherent impls of `f32`/`f64::mul_add` available.  Please change these tests to explicitly call `MulAdd::mul_add(m, x, b)`, and then this test will need a `cfg` gate.

PR #61, opened 2018-04-10
Use constant for 180/π in f32::to_degrees

vks:
The current `f32::to_degrees` implementation uses a division to
calculate 180/π, which causes a loss of precision. Using a constant is
still not perfect (implementing a maximally-precise algorithm would come
with a high performance cost), but improves precision with a minimal
change.

This is a backport from [`std`].

[`std`]: https://github.com/rust-lang/rust/pull/47919/commits/e34c31bf02eb0f0ff4dd43ae72e0eae53f2ac519

### DIFF
--- a/src/float.rs
+++ b/src/float.rs
@@ -2015,6 +2015,9 @@ mod tests {
     }
 
     #[test]
+    // This fails with the forwarded `std` implementation in Rust 1.8.
+    // To avoid the failure, the test is limited to `no_std` builds.
+    #[cfg(not(feature = "std"))]
     fn to_degrees_rounding() {
         assert_eq!(1_f32.to_degrees(), 57.2957795130823208767981548141051703);
     }
--- a/src/float.rs
+++ b/src/float.rs
@@ -2019,6 +2019,8 @@ mod tests {
     // To avoid the failure, the test is limited to `no_std` builds.
     #[cfg(not(feature = "std"))]
     fn to_degrees_rounding() {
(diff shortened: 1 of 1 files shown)


cuviper:
Just for reference, we also discussed the implication that some inputs are now less accurate, in rust-lang/rust#48617.  But the rounding is more correct in general, and here we forward to `std` when possible anyway.

src/float.rs
```diff
@@ -2008,4 +2013,9 @@ mod tests {
             assert!((Float::to_radians(deg) - rad).abs() < 1e-5);
         }
     }
+
+    #[test]
+    fn to_degrees_rounding() {
+        assert_eq!(1_f32.to_degrees(), 57.2957795130823208767981548141051703);
```
cuviper: When we forward to `std`, this fails on Rust 1.8 since it lacks rust-lang/rust#47919.  Maybe just limit this test to no-std builds? (with a comment explaining this reason)

cuviper:
Strange, 1.8 still failed!

```
---- float::tests::to_degrees_rounding stdout ----
	thread 'float::tests::to_degrees_rounding' panicked at 'assertion failed: `(left == right)` (left: `57.295776`, right: `57.29578`)', src/float.rs:2022
```

cuviper:
Solved by explicitly calling `FloatCore` in the test.

bors r+

vks:
Thanks for the fix! I forgot about this PR...

PR #58, opened 2018-04-06
Add `CheckedRem` and `CheckedNeg`

LEXUGE:
Continue form #56 
Add `CheckedRem` and `CheckedNeg` traits
I've only changed /src/ops/checked.rs, other changes are came from rustfmt.

PR #63, opened 2018-04-13
Add CheckedRem and CheckedNeg

LEXUGE:
Continue from #58 
I've alreadyremoved all the formats.


cuviper:
Thanks -- I also added some changes myself.

bors r+

Issue #47, opened 2018-03-05
Tracking issue for potential breaking changes


cuviper:
num-traits is somewhat of a _de facto_ stable crate.  Since these traits are often found in the public API of other crates, we really want to avoid the sort of ecosystem split that a breaking-change semver bump might cause.

Nevertheless, it's a good idea to track what breaking changes we **might** want to make someday, now found in the list below.  Links are included to postponed issues, which often have more details.  They will be closed in the meantime, but discussion can continue.

- Remove default `epsilon` implementation for `Float` #4
- Num::FromStrRadixErr should have Trait bound #6
- More features for `Float` #8
- All bit twiddling functions in num::traits::PrimInt should take or return usize instead of u32 #14
- `std::num::Checked{Add,Sub,Mul,Div}` should have type parameters #15
- Use `Mul` by reference in `pow` #18
- Make the Float trait depend on FloatConst. #20
- Add Default, Display, and Debug trait bounds to PrimInt #44
- Easy conversions between `Signed` and `Unsigned` #46

kotauskas:
I don't get why #15 is considered breaking. `Rhs=Self` should keep it compatible.

cuviper:
Quoting #15:
> The corresponding `std::ops` traits have both the type of right hand side operand and the type of result, while `Checked*` traits do not (and always assume that both operands and result have the same type).

I vaguely recall seeing some cases where type defaults don't fully prevent breaking changes, but I'm not sure off-hand. The result type is definitely a problem though, because [associated type defaults](https://github.com/rust-lang/rust/issues/29661) are still unstable.

kotauskas:
IMO the `Result` type should not be added just yet and still be postponed for 1.0 or future 0.x releases, and `Rhs=Self`, something that I really need for [`bigbit`](https://crates.io/crates/bitbit) now, added in 0.2.x. Just sitting down and brainstorming this, I have no idea how `Rhs=Self` could ever break anything.

cuviper:
Here's the breaking case I was thinking of: https://github.com/rust-lang/rust/pull/69454#issuecomment-591877207

newpavlov:
How about introducing to the conversion traits `try_from` method in addition to `from` and make the latter return value instead of `Option` (when conversion can not be done it will panic)? It's quite annoying to write `F::from(42.0).unwrap()` and it makes resulting code containing many constants like this really noisy.

kotauskas:
[...]

PR #65, opened 2018-05-02
impl Bounded for Reverse

andersk:
[`Reverse`](https://doc.rust-lang.org/std/cmp/struct.Reverse.html) is stable as of Rust 1.19.

### DIFF
--- a/src/bounds.rs
+++ b/src/bounds.rs
@@ -1,6 +1,7 @@
 use core::{usize, u8, u16, u32, u64};
 use core::{isize, i8, i16, i32, i64};
 use core::{f32, f64};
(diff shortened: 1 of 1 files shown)


cuviper:
> stable as of Rust 1.19.

That's newer than the 1.8 we support.

It's possible to add new impls behind a feature flag, like #60 is doing for `i128`, but `Bounded for Reverse` seems like a pretty niche thing to justify a feature.

cuviper:
I've add this to the tracking issue #47 for breaking changes.  I'm going to close this for now, but if you want to make a case for adding this with a conditional feature, we can discuss that.

PR #70, opened 2018-05-20
Clarify in the docs that `mul_add` is not always faster.

frewsxcv:
More info:

- https://github.com/rust-lang/rust/issues/49842
- https://github.com/rust-lang/rust/pull/50572


cuviper:
Thanks!

bors r+

Issue #77, opened 2018-07-18
pow(0, 0) returns 1.


meltinglava:
`pow(0, 0)` should be undefined. Should probably panic in `pow(0, 0)`, and return `None` in `checked_pow(0, 0)`.

cuviper:
Documented in #79.

PR #78, opened 2018-07-18
Pow(0, 0) no longer returnes 1

meltinglava:


src/pow.rs
```diff
@@ -182,9 +183,13 @@ mod float_impls {
 /// assert_eq!(pow(6u8, 3), 216);
 /// ```
 #[inline]
-pub fn pow<T: Clone + One + Mul<T, Output = T>>(mut base: T, mut exp: usize) -> T {
+pub fn pow<T: Clone + Zero + One + Mul<T, Output = T>>(mut base: T, mut exp: usize) -> T {
     if exp == 0 {
-        return T::one();
+        return if base.is_zero() {
+            panic!("0⁰ is undefined")
```
hauleth: I would restrain myself from using Unicode in panic messages as it can be problematic on platforms that do not support Unicode in CLI (aka Windows).

src/pow.rs
```diff
@@ -182,9 +183,13 @@ mod float_impls {
 /// assert_eq!(pow(6u8, 3), 216);
 /// ```
 #[inline]
-pub fn pow<T: Clone + One + Mul<T, Output = T>>(mut base: T, mut exp: usize) -> T {
+pub fn pow<T: Clone + Zero + One + Mul<T, Output = T>>(mut base: T, mut exp: usize) -> T {
```
hauleth: I believe that this is breaking change, so we should think if this is really only way.

src/pow.rs
meltinglava: do we want: 
* 0^0
* pow(0, 0)
* any other?

hauleth: I would go with `pow(0, 0)` as this is what user wrote. `0^0` is common syntax but in this case can be confused with XOR operator.

cuviper:
cc #77 for the cross-reference

Many other languages return 1: https://rosettacode.org/wiki/Zero_to_the_zero_power

I think we should leave the current behavior alone and just document this edge case.  Since we can't even check for this case without adding `Zero` in a breaking change, returning `1` is a reasonable compromise with tons of precedent.

cuviper:
Closing in favor of #79.

PR #79, opened 2018-07-21
Updated documentation to note the pow(0, 0) case.

meltinglava:
Ref #78

### DIFF
--- a/src/pow.rs
+++ b/src/pow.rs
@@ -173,7 +173,7 @@ mod float_impls {
 
 /// Raises a value to the power of exp, using exponentiation by squaring.
 ///
-/// Note that `0⁰` (`pow(0, 0)`) returnes `1`. Mathematicly this is undefined.
+/// Note that `0⁰` (`pow(0, 0)`) returnes `1`. Mathematically this is undefined.
 ///
 /// # Example
 ///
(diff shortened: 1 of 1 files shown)


src/pow.rs
```diff
@@ -173,13 +173,16 @@ mod float_impls {
 
 /// Raises a value to the power of exp, using exponentiation by squaring.
 ///
+/// Note that `0⁰` (`pow(0, 0)`) returnes `1`. Mathematicly this is undefined.
```
cuviper: spelling: should be "mathematically"

src/pow.rs
```diff
@@ -208,6 +211,8 @@ pub fn pow<T: Clone + One + Mul<T, Output = T>>(mut base: T, mut exp: usize) ->
 
 /// Raises a value to the power of exp, returning `None` if an overflow occurred.
 ///
+/// Note that `0⁰` (`checked_pow(0, 0)`) returnes `Some(1)`. Mathematicly this is undefined.
```
cuviper: *mathematically

hauleth:
Quoting prof Knuth:

> On the other hand, Cauchy had good reason to consider 0^0 as an undefined limiting form, in the sense that the limiting value of f(x)^g(x) is not known a priori when f(x) and g(x) approach 0 independently. In this much stronger sense, the value of 0^0 is less defined than, say, the value of 0 + 0. Both Cauchy and Libri were right, but Libri and his defenders did not understand why truth was on their side.

https://arxiv.org/abs/math/9205211

cuviper:
@hauleth It's not clear to me if you're quoting that in favor or against, or just representing the debate.

Anyway, commenting the existing behavior is harmless.

bors r+

hauleth:
@cuviper my comment was in favour of current behaviour as I think it is more reasonable solution in case of computation. `0^0` is "mathematically undefined" when we are working in area of mathematical analysis, and in such case `num` crate isn't that much helpful. We could quote this paper as a reasoning for current behaviour and why we think it is acceptable for us.

PR #81, opened 2018-09-02
Add wrapping shifts

edmccard:
Add traits `WrappingShl` and `WrappingShr` corresponding to the
standard library `wrapping_shl` and `wrapping_shr` methods. Implement
the trait on all primitive integer types as well as on `Wrapping`.


cuviper:
Looks great, thanks!

bors r+

PR #88, opened 2018-09-29
Add safe cast operation traits

npmccallum:
This commit adds safe cast operation traits which each have only one of
the properties from a naked cast. This allows the clear definition of
specific properties in a generic context.

All methods defined in this commit are zero-cost abstractions.

### DIFF
--- a/src/cast/safe/cast.rs
+++ b/src/cast/safe/cast.rs
@@ -1,0 +1,103 @@
+/// Cast from one numeric type to another.
+///
+/// The `CastFrom<T>` trait is similar to `std::convert::From<T>`. However,
+/// while `std::convert::From<T>` performs a logical conversion, `CastFrom<T>`
+/// performs a bitwise cast from one number to another; possibly of a different
+/// size, signedness or even numeric type.
+///
+/// Unless you really know what you are doing, you probably don't want this
+/// trait. Instead, you should check out the following traits:
+///
+///   * `GrowFrom<T>`
+///   * `TrimFrom<T>`
+///   * `SignCast`
+pub trait CastFrom<T> {
+    fn cast(value: T) -> Self;
+}
+
+/// Cast from one numeric type to another.
+///
+/// The `CastInto<T>` trait is similar to `std::convert::Into<T>`. However,
+/// while `std::convert::Into<T>` performs a logical conversion, `CastInto<T>`
+/// performs a bitwise cast from one number to another; possibly of a different
+/// size, signedness or even numeric type.
+///
+/// Unless you really know what you are doing, you probably don't want this
+/// trait. Instead, you should check out the following traits:
+///
+///   * `GrowInto<T>`
+///   * `TrimInto<T>`
+///   * `SignCast`
+pub trait CastInto<T> {
+    fn cast(self) -> T;
+}
+
+// CastFrom implies CastInto
+impl<T, U> CastInto<U> for T where U: CastFrom<T>
+{
+    #[inline]
+    #[must_use]
+    fn cast(self) -> U {
+        U::cast(self)
+    }
+}
+
+// CastFrom (and thus CastInto) is reflexive
+impl<T> CastFrom<T> for T {
+    #[inline]
+    #[must_use]
+    fn cast(t: T) -> T { t }
+}
+
+macro_rules! cast_impl {
+    ($from:ty > $into:ty) => (
+        impl CastFrom<$from> for $into {
+            #[inline]
+            #[must_use]
+            fn cast(value: $from) -> $into {
+                value as $into
+            }
+        }
+    );
+
+    (i128 => $into:ty) => (
+        #[cfg(has_i128)]
+        cast_impl! { i128 > $into }
+    );
+
+    (u128 => $into:ty) => (
+        #[cfg(has_i128)]
+        cast_impl! { u128 > $into }
+    );
+
+    ($from:ty => i128) => (
+        #[cfg(has_i128)]
+        cast_impl! { $from > i128 }
+    );
+
+    ($from:ty => u128) => (
+        #[cfg(has_i128)]
+        cast_impl! { $from > u128 }
+    );
+
+    ($from:ty => $into:ty) => (
+        cast_impl! { $from > $into }
+    );
+
+    ($kind:ty, $($next:ty),+) => (
+        $(
+            cast_impl! { $kind => $next }
+            cast_impl! { $next => $kind }
+        )+
+
+        cast_impl! { $($next),+ }
+    );
+
+    ($kind:ty) => ();
+}
+
+cast_impl! {
+    usize, u128, u64, u32, u16, u8,
+    isize, i128, i64, i32, i16, i8,
+    f32, f64
+}
--- a/src/cast/safe/grow.rs
+++ b/src/cast/safe/grow.rs
@@ -1,0 +1,113 @@
+/// Cast from a smaller number to a larger one without changing sign.
+///
+/// The `GrowFrom<T>` trait is similar to `std::convert::From<T>`. However,
+/// while `std::convert::From<T>` performs a logical conversion, `GrowFrom<T>`
+/// performs a bitwise cast from a number to a number of the same sign but
+/// of a possibly larger size. `GrowFrom<T>` will **never** decrease the size
+/// of a number or change from an integer of one signedness to the other.
+pub trait GrowFrom<T> {
+    #[inline]
+    #[must_use]
+    fn grow(value: T) -> Self;
+}
+
+/// Cast from a smaller number to a larger one without changing sign.
+///
+/// The `GrowInto<T>` trait is similar to `std::convert::Into<T>`. However,
+/// while `std::convert::Into<T>` performs a logical conversion, `GrowInto<T>`
+/// performs a bitwise cast from a number to a number of the same sign but
+/// of a possibly larger size. `GrowInto<T>` will **never** decrease the size
+/// of a number or change from an integer of one signedness to the other.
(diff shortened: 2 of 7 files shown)


src/cast/safe/cast.rs
```diff
@@ -0,0 +1,103 @@
+/// Cast from one numeric type to another.
+///
+/// The `CastFrom<T>` trait is similar to `std::convert::From<T>`. However,
+/// while `std::convert::From<T>` performs a logical conversion, `CastFrom<T>`
+/// performs a bitwise cast from one number to another; possibly of a different
+/// size, signedness or even numeric type.
```
cuviper: I don't understand why you call this a bitwise cast -- that doesn't make any sense to me for types of different sizes.  Really you're just casting with `as`, which is also a "logical" (if sometimes lossy) conversion, especially when floating point is involved.

Overall, I don't see how `CastFrom` has any meaningful difference from `AsPrimitive`.

src/cast/safe/grow.rs
```diff
@@ -0,0 +1,113 @@
+/// Cast from a smaller number to a larger one without changing sign.
+///
+/// The `GrowFrom<T>` trait is similar to `std::convert::From<T>`. However,
+/// while `std::convert::From<T>` performs a logical conversion, `GrowFrom<T>`
+/// performs a bitwise cast from a number to a number of the same sign but
+/// of a possibly larger size. `GrowFrom<T>` will **never** decrease the size
+/// of a number or change from an integer of one signedness to the other.
```
cuviper: Calling this "bitwise" is again weird.  Otherwise, the difference from `From` is just that you're preserving sign, but I wonder how useful that really is in generic code?  If you care to remain signed, you can add a constraint with the `Signed` trait, or likewise with the `Unsigned` trait.  But if you're in generic code that doesn't know whether it's signed or unsigned, I'm not sure that I see it being useful to assert that the signedness won't change.

src/cast/safe/sign.rs
```diff
@@ -0,0 +1,232 @@
...
+/// `SignCast` exposes the cast operator in generic contexts. It is used to
+/// perform a signed cast between integers of the same size. The `SignCast`
+/// trait will **never** change the size of the integer.
+pub trait SignCast {
+    type Unsigned;
+    type Signed;
+
+    fn unsigned(self) -> Self::Unsigned;
+    fn signed(self) -> Self::Signed;
+}
```
cuviper: I do like this trait -- I have often wanted this functionality.

Oddly, this one actually is a bitwise conversion, but you didn't call it such...

I think we should strengthen the associated types using the sign traits:

```rust
    type Unsigned: Unsigned;
    type Signed: Signed;
```

src/cast/safe/size.rs
```diff
@@ -0,0 +1,103 @@
...
+///
+/// Unless you really know what you are doing, you probably don't want this
+/// trait. Instead, you should check out the following traits:
+///
+///   * `GrowFrom<T>`
+///   * `TrimFrom<T>`
+///   * `SignCast`
+pub trait SizeFrom<T> {
+    fn size(value: T) -> Self;
+}
```
cuviper: Like my comments on `GrowFrom`, I don't see the value of maintaining an unknown signedness.  If you do know the sign you're dealing with, then constraints on `AsPrimitive` and `Signed`/`Unsigned` should do the same job.

src/cast/safe/trim.rs
```diff
@@ -0,0 +1,109 @@
+/// Cast from a larger number to a smaller one without changing sign.
+///
+/// The `TrimFrom<T>` trait is similar to `std::convert::From<T>`. However,
+/// while `std::convert::From<T>` performs a logical conversion, `TrimFrom<T>`
+/// performs a bitwise cast from a number to a number of the same sign but
+/// of a possibly smaller size. `TrimFrom<T>` will **never** increase the size
+/// of a number or change from an integer of one signedness to the other.
```
cuviper: Again, the value of unknown signedness is not clear to me.

And once you've hedged that this is only *possibly* a smaller size, how is this better than `AsPrimitive`?

src/cast/safe/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+//! not implemented for `u8`).
+//!
+//! ```rust
+//! use num_traits::cast::safe::{SignCast, TrimInto};
+//!
+//! const MASK: u16 = 0b0000111111111111;
+//!
+//! fn low12<T: TrimInto<u16>, C: SignCast<Unsigned=T>>(x: C) -> u16 {
+//!     x.unsigned().trim() & MASK
+//! }
```
cuviper: Since this is your only concrete usage example, how is it much better than:

```rust
fn low12<T: AsPrimitive<u16>, C: SignCast<Unsigned=T>>(x: C) -> u16 {
    x.unsigned().as_() & MASK
}
```

Or personally, I'd write the bounds like this:

```rust
fn low12<T: SignCast>(x: T) -> u16
where
    T::Unsigned: AsPrimitive<u16>,
{
    x.unsigned().as_() & MASK
}
```

This doesn't exclude `u8` input, but is that really necessary?

cuviper:
Regarding the CI failure, you need `#[cfg(has_i128)]` on 128-bit functionality.

src/cast/safe/cast.rs
npmccallum: Bitwise was the wrong word. Sorry. Late night. :)

`CastFrom` is a stylistic cleanup of AsPrimitive to match the other new traits if they are merged. The idea is to have identical syntax across all the conversion types.

src/cast/safe/grow.rs
npmccallum: That is true, but that behavior is implicit. A big part of these traits is turning implicit behavior into explicit behavior. This improves readability and reduces run time errors.

src/cast/safe/sign.rs
npmccallum: That makes sense to me.

src/cast/safe/size.rs
npmccallum: See my comment above about readability. When someone reads code and sees `as` they don't know what actually happened. This is about making all behaviors of the cast as explicit as possible.

src/cast/safe/mod.rs
npmccallum: Yes. Think of a nuclear silo. Pretend there is a flag that controls some aspect of launching the rocket in bit 11. Do you want people passing a small integer into the function and getting unexpected behavior?

I'd rather the interface clearly define: 16 bits are the minimum. The compiler will fail otherwise.

npmccallum:
> Regarding the CI failure, you need `#[cfg(has_i128)]` on 128-bit functionality.

Is this still needed now that `i128` is stable?

npmccallum:
> And overall, grouping these in a safe module has a weird implication for the rest of the cast module.

Considering that I'm trying to move implicit behavior to explicit behavior, I think safe is pretty reasonable. Honestly, I'd prefer that these interfaces (properly reviewed) replace the rest of the cast module.

> The traits like FromPrimitive and ToPrimitive are explicitly safe conversions too, just fallible.

**At runtime.** One of the whole benefits of Rust is catching errors at compile time. And casts are a pretty regular source of bugs.

cuviper:
> Yes. Think of a nuclear silo. Pretend there is a flag that controls some aspect of launching the rocket in bit 11. Do you want people passing a small integer into the function and getting unexpected behavior?
> 
> I'd rather the interface clearly define: 16 bits are the minimum. The compiler will fail otherwise.

This is a scenario where I would eschew numeric conversions altogether, and instead insist on a strong newtype wrapper, probably even based on [bitflags](https://crates.io/crates/bitflags).

> > Regarding the CI failure, you need `#[cfg(has_i128)]` on 128-bit functionality.
> 
> Is this still needed now that `i128` is stable?

Yes, I maintain strong backwards compatibility here.  If I were to require a new rustc in `num-traits`, that would *force* every dependent crate to increase their requirement too.  I choose to remain conservative about this.

> **At runtime.** One of the whole benefits of Rust is catching errors at compile time. And casts are a pretty regular source of bugs.

As much as possible is at compile time, but some things are necessarily at runtime, like integer overflow and bounds checking.  For numeric conversions, we have:
- static lossless: `std::convert::{From,Into}`
- static lossy: `as`, `num_traits::cast::AsPrimitive`
- runtime checked: (unstable) `std::convert::{TryFrom,TryInto}`, `num_traits::cast::{FromPrimitive,ToPrimitive,NumCast}`

Your traits add grow/trim/sign dimensions to the static case, but I'm not yet convinced that those are really any *safer* than the simple lossless/lossy distinction.

> Considering that I'm trying to move implicit behavior to explicit behavior, I think safe is pretty reasonable. Honestly, I'd prefer that these interfaces (properly reviewed) replace the rest of the cast module.

I think what I'm coming around to, is that I don't really agree that these are *better* than the existing cast module.  Even if I did, given that the others can't go away without a breaking change (which I'm avoiding as long as I can), we'd end up with a redundancy here that will make it more confusing for the users.

However, there is always room for a new addition on crates.io!  I was trying to recall some other existing crates in this space, and I believe it was [conv](https://crates.io/crates/conv) that I've seen before.  I also found [cast](https://crates.io/crates/cast).  These are both fallible (runtime checked) though, unlike your static guarantees.

cuviper:
I thought you might also be interested in this thread:
https://internals.rust-lang.org/t/idea-foo-bar-type-conversion-operator/8542

cuviper:
I'm going to close this, since we disagree on design and the discussion stalled. You can re-open or start a new issue if you'd like to revisit this.

Issue #54, opened 2018-03-19
Associated constants for One and Zero


kvark:
Any reason we can't have `ONE` and `ZERO` implemented as constants rather than methods? Given that `num_traits` is a low level crate, it effectively forbids any higher level constructs with associated constants atm.

kvark:
Unless there are objections, I'm giving this a try (affecting way more traits than I expected, but still).

cuviper:
It would be a breaking change to both users and implementors, which we strongly avoid, and associated constants require a newer rustc than our conservative 1.8.  Also, some types which implement these traits can't have constant values at all, like `BigInt`.

kvark:
@cuviper  thanks for the answer!

> It would be a breaking change to both users and implementors, which we strongly avoid,

The way I see it is adding the associated consts first, using them internally where possible, marking the old methods with `#[deprecated]` and then removing the old ones in 0.3. So I don't think this is an issue.

> associated constants require a newer rustc than our conservative 1.8

I'm curious about this. Why 1.8 in particular? What are the guidelines/schedule for updating this base requirement?

> some types which implement these traits can't have constant values at all, like BigInt

This is truly unfortunate, and it's a real blocker.

kvark:
Following up on  the last one - there does appear to be some light at the end of the tunnel, see https://github.com/rust-num/num-bigint/issues/36

cuviper:
> The way I see it is adding the associated consts first

Even that is a breaking change for third-party implementations of the traits.  The only non-breaking way to expand a trait is with a default implementation of new items, which we can't do here.

> I'm curious about this. Why 1.8 in particular? What are the guidelines/schedule for updating this base requirement?

It used to be 1.0, then other factors forced us to raise it to 1.8 -- rust-num/num#257.  I treat raising the minimum rustc as a breaking change, but the community has mixed opinions on this -- rust-lang-nursery/api-guidelines#123.  I don't have a schedule, but I just keep it conservative, since num-traits is a common dependency, often even in other crates' public API.  Increasing num-traits' minimum rustc will force that decision on all of its dependents too.

> Following up on the last one - there does appear to be some light at the end of the tunnel, see rust-num/num-bigint#36

Maybe, but this is just one example.  We don't control all possible implementations of these traits.

cuviper:
Of course, nothing stops you from implementing your own `One` trait with different trade-offs.  The ubiquity of num-traits is a benefit to its use, but also constrains its ability to change.

[...]

PR #91, opened 2018-10-17
Add ConstZero and ConstOne traits

gluyas:
As described in [issue #54](https://github.com/rust-num/num-traits/issues/54#issuecomment-401425344). 
Works much the same way as `i128` support: probes the compiler in the build script, and can be made a requirement by enabling the `associated_consts` crate feature. 

There are several other traits in the crate which have similar "pseudo-consts", such as `Bounded` and `Float`, which should also eventually make use of associated constants through additional traits. However, `Float` in particular is not so straightforward; it already has its own `FloatConst` subtrait, which would require either a breaking change or a different naming convention to be extended, among other things.

A `NumConst` trait also seems like a natural fit; most types will implement either all or none of these const-y traits. However, it could be worth holding off on this, in case [`const fn`](https://github.com/rust-lang/rust/issues/24111) ends up offering support for trait methods. Something like `NumOps` with compile-time evaluation would be a very powerful feature for generic code.

My suggestion is that #54 be closed and a new issue opened to discuss a broader implementation of associated constants in the crate. It's also likely that the discussion would affect the contents of this pull, so I think it's probably worth withholding these changes until then.

Very interested to hear others' thoughts on this!


cuviper:
> Works much the same way as `i128` support: probes the compiler in the build script, and can be made a requirement by enabling the `associated_consts` crate feature.

I'm not sure I want an explicit feature for this.  That was kind of an accident for `i128`, as I started with just a feature, and then I quickly got user feedback that auto-detection would be more useful to dependent crates.  In this new case, I think we can just leave it to the build script alone.

I recently started an [`autocfg` crate](https://crates.io/crates/autocfg) which I want to start using throughout `num` crates, but it only has path/type/trait probing right now.  I should probably add support for arbitrary expressions for cases like you've written here.

> There are several other traits in the crate which have similar "pseudo-consts", such as `Bounded` and `Float`, which should also eventually make use of associated constants through additional traits. However, `Float` in particular is not so straightforward; it already has its own `FloatConst` subtrait, which would require either a breaking change or a different naming convention to be extended, among other things.

Maybe `ConstFoo` would be a better naming convention?  It seems to read a little more directly to have `ConstZero` provide `const ZERO`, for instance.  Then we kind of dodge the `Float` issue by having the existing `FloatConst` and a new `ConstFloat` -- that will be a little confusing, but acceptable IMO.

> A `NumConst` trait also seems like a natural fit; most types will implement either all or none of these const-y traits.

A direct parallel to `Num` would only include `ZERO` and `ONE`.  I'm not sure what others would fit "most" types.  Integers won't have the floating point constants, of course. `BigInt` and `BigUint` can't have bounded constants, since they're unbounded, but then they can't really implement `ZERO` or `ONE` either until there's a way to have a constant backing `Vec`.

cuviper:
I'm still interested, if you care to discuss my earlier reply. Otherwise, I may reimplement this separately.

gluyas:
Hi! Sorry for never getting back to this...

All the changes you suggested sound good and I'd be happy to implement them.

One of my concerns was forward-compatibility for the eventual implementation of compile-time math expressions. My intuition was that we would need another trait with `const fn` method implementations, `ConstNumOps`, which would fit in with `ConstNum` (which I was unclear about in my previous post).

But it looks like the [current RFC proposal](https://github.com/rust-lang/rfcs/pull/2632) will make these library-defined const-y traits unnecessary. It adds a `const Trait` syntax that makes all the trait methods in a specific implementation `const`; exactly what we are doing here.

But that's probably far enough away for these traits (and maybe a few others?) to be useful as a stop-gap until then. Of course the RFC could change so I think it's worth only making other const-t traits when there is a concrete use-case with the current features in Rust. I think `Zero`, `One`, and `Bounded` are good candidates for that.

PS: Should I rebase this branch so it's up to date? Or is it cleaner to do it in the merge commit? Thanks!

gluyas:
I went ahead and rebased this onto `rust-num:master` (and tidied up some of the commits) so this shouldn't have any problems being merged. I also added an implementation for a `ConstBounded` trait, I figure it's not hard to revert if it you don't think it's a good fit.

Coder-256:
Just wondering, what is the benefit of associated constants over for example, the soon-coming `impl const Zero`? I would argue that the latter is better and has the same functionality without needing a separate trait. Plus it requires no changes: people can use new versions of rust with those capabilities while the crate can still support old versions (I could be wrong about that)

cuviper:
@gluyas 
> But it looks like the [current RFC proposal](https://github.com/rust-lang/rfcs/pull/2632) will make these library-defined const-y traits unnecessary. It adds a `const Trait` syntax that makes all the trait methods in a specific implementation `const`; exactly what we are doing here.

@Coder-256 
> Just wondering, what is the benefit of associated constants over for example, the soon-coming `impl const Zero`? I would argue that the latter is better and has the same functionality without needing a separate trait.

I wasn't aware of that proposal -- somehow I missed when @gluyas first linked it.

So yes, that looks great! The only disadvantage will be waiting for this support to reach a stable release of `rustc`, but long term this does look like a better way to approach these constants. Since that RFC is now in fcp-merge status, I'm inclined to leave it to that, rather than expanding `num-traits` with something that will be inferior. It seems better to suggest that folks can use their own local traits for `ConstZero`/etc. as a stop-gap, rather than being stuck in our API. OK?

> Plus it requires no changes: people can use new versions of rust with those capabilities while the crate can still support old versions

We'll probably want `impl const` on a bunch of the local trait implementations for primitives, but the traits themselves shouldn't need any change.

gluyas:
Seems good. Closing this favour of waiting for language support.
Should also resolve #54.

PR #96, opened 2019-01-10
Add Modular Inverse Trait

goldenMetteyya:


cuviper:
Is this a work in progress? I would at least expect an implementation of the trait for the primitive types.

Since this is specific to integers, it probably belongs in the `num-integer` crate. You can see that rust-num/num-integer#10 even started an inverse function.

cuviper:
Please take this to `num-integer` if you want to work further.

Issue #75, opened 2018-07-16
Use the libm crate for a full Float on no_std


cuviper:
Currently, `num_traits::Float` is only available with `std`, and `no_std` only has the more limited `FloatCore` without all of transcendental functions like `sqrt`, `sin`, etc.  The new `libm` crate might be the way out of this, as an optional dependency to get `Float` on `no_std` too!

There's also `Real` and the signed and floating-point `Pow` (mapping to `powi` and `powf`) which all require `std` now, and should be feasible with `libm`.  I'm not sure about `MulAdd` and `MulAddAssign` though.

vks:
`libm` is not ready yet, it panics with overflows in debug mode.

cuviper:
Sure, we can hold off until it's more ready -- I'm not in any rush.  It could still be useful to do early exploration though, to identify any other shortcomings that `libm` might want to address.

malobre:
Since #99 was merged shouldn't this be closed ?

korken89:
Hi, please make a release including #99 this is a key component for a library I'm writing and I would love to have this!

CC @cuviper

cuviper:
I will try to get to it this week.

CryZe:
What's the status of this? It's been two weeks (sorry for maybe pinging this early, but this is my last blocker).

korken89:
Same here, this is the only thing that is missing for me as well

cuviper:
Sorry for the delay -- num-traits 0.2.9 is now published!

PR #99, opened 2019-02-03
Revive Float+Real in no_std thanks to libm

yoanlcq:
Greetings,

This is a hopeful fix for #75.  
Basically: Add `libm` as an optional dependency, and handle three possible cases depending on which features are enabled:
- std and libm: std is used;
- std and not libm: std is used;
- libm and not std: libm and FloatCore are used.

It was briefly mentioned that `libm` wasn't ready yet, but this was months ago, and I believe it is better not to wait for too long.  
If anything, bugs in `libm` should be fixed in `libm`; `num-traits` is only delegating its implementations to it; not to mention that the more `libm` is used, the likelier issues are to be found and hopefully fixed.

Thanks in advance!

### DIFF
--- a/src/real.rs
+++ b/src/real.rs
@@ -1,9 +1,8 @@
+#![cfg(any(feature = "std", feature = "libm"))]
+
 use core::ops::Neg;
 
-use {Num, NumCast};
-
-#[cfg(any(feature = "std", feature = "libm"))]
-use Float;
+use {Float, Num, NumCast};
 
 // NOTE: These doctests have the same issue as those in src/float.rs.
 // They're testing the inherent methods directly, and not those of `Real`.
@@ -14,14 +13,11 @@ use Float;
 /// See [this Wikipedia article](https://en.wikipedia.org/wiki/Real_data_type)
 /// for a list of data types that could meaningfully implement this trait.
 ///
-/// This trait is always available, however it requires either `std` or `libm`
-/// in order for `f32` and `f64` to implement it.
-#[cfg(any(feature = "std", feature = "libm"))]
+/// This trait is only available with the `std` feature, or with the `libm` feature otherwise.
 pub trait Real: Num + Copy + NumCast + PartialOrd + Neg<Output = Self> {
     /// Returns the smallest finite value that this type can represent.
     ///
-    #[cfg_attr(any(feature = "std", feature = "libm"), doc = " ```")]
-    #[cfg_attr(not(any(feature = "std", feature = "libm")), doc = " ```ignore")]
+    /// ```
     /// use num_traits::real::Real;
     /// use std::f64;
     ///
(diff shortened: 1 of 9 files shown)


yoanlcq:
That should do; hopefully I didn't forget anything :)

yoanlcq:
CI fails because of overflow issues with `libm`. So I guess this is blocked by [https://github.com/rust-lang-nursery/libm/issues/4](https://github.com/rust-lang-nursery/libm/issues/4).

src/real.rs
```diff
@@ -777,6 +828,7 @@ pub trait Real: Num + Copy + NumCast + PartialOrd + Neg<Output = Self> {
     fn atanh(self) -> Self;
 }
 
+#[cfg(any(feature = "std", feature = "libm"))]
 impl<T: Float> Real for T {
```
cuviper: We need to guard the trait definition in the same way. Features must additive, but adding a blanket impl to a trait is a breaking change. That's not an issue if the trait never exists without the blanket impl.

cuviper:
Thanks for doing this - it looks great! But yes, the `libm` overflows are a problem...

Schultzer:
Hi @yoanlcq and @cuviper a new version of libm has been released and it seams that the overflow problem with the tests in debug mode has been resolved any chance that we can move forward with this? :)

korken89:
Hi, any progress on this? cc @cuviper

CryZe:
I'd like to see this move forward too, as there's quite a few crates that I'd like to send PRs to for no_std support that depend on this.

yoanlcq:
I think it's pretty much ready now; the only problem seems to be that `libm` doesn't compile on some older Rust versions supported by `num-traits`.

korken89:
Awesome, thanks for doing the work!

src/real.rs
yoanlcq: This is now normally fixed [in this commit](https://github.com/rust-num/num-traits/pull/99/commits/583f2c71a3ff023de0b5d12714c95d21de5c9f6e).

cuviper:
> the only problem seems to be that `libm` doesn't compile on some older Rust versions supported by `num-traits`.

I'm OK with that -- this is new functionality for `no_std` users, so it's fine for this to have less compatibility. Just document that it's limited by the MSRV of the `libm` crate, and adjust the CI script accordingly (like `i128`).

yoanlcq:
That should do, I hope :) Sorry for the delay.

CryZe:
What's the status of this? This would unblock a lot of PRs / releases for other crates.

korken89:
It should be ready for merge, so I think we are waiting for the maintainers

cuviper:
I rebased and tweaked a few things, e.g. `Real` doesn't need all those doc-test guards now that we're guarding the trait entirely. I think we're ready, but I'd appreciate if anyone else takes another look.

yoanlcq:
The single guard in `mod real` is so obviously better, that I wonder why I didn't do it this way in the first place...

Anyway, I couldn't find anything "wrong" in these last changes. I would pretty confidently say "ship it"!

cuviper:
> The single guard in `mod real` is so obviously better, that I wonder why I didn't do it this way in the first place...

I think because you had the trait itself unguarded at first, only the impl depending on std/libm. So all those doctests had to be guarded whether there was any useful type available to test.

bors r+

PR #100, opened 2019-02-09
Added real and imaginary parts methods

YakoYakoYokuYoku:
I added them in purpose to match my num-complex pull request.


YakoYakoYokuYoku:
TODO: Fix this implementation :weary:

cuviper:
What is your motivating use case for these trait methods? Are you writing generic code that doesn't know whether it has an actual `Complex` type?

I think this would be better off as a new trait. The default impls you've added to `Float` are wrong for any type which is not purely real, and we have no way of knowing this from here. Plus, these methods are not really specific to floating point anyway -- for instance Gaussian integers are complex too.

YakoYakoYokuYoku:
Closing in favor of #101.

PR #101, opened 2019-02-21
Mirror of num-complex issue #2

YakoYakoYokuYoku:
Details are [here](https://github.com/rust-num/num-complex/pull/40).


YakoYakoYokuYoku:
I think the parts where is failing is in:
`Float where Self: // --snip--`
and in:
`CommonFloat where Self: // --snip--`

I think using `Float: // --snip--` is correct.

YakoYakoYokuYoku:
@cuviper sorry to bother you with this kind of stuff and for the inactivity, but i couldn't make the commits in this PR to work (in my horse-shoe crate it does) and i don't know why. If you can help me with this that would be nice.

PR #103, opened 2019-02-28
convert between integer and byte array

flier:
After Rust 1.32, the function `to_<endian>_bytes` and `from_<endian>_bytes` has [stabilizations](https://blog.rust-lang.org/2019/01/17/Rust-1.32.0.html#library-stabilizations).

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -15,6 +15,8 @@ fn main() {
     if env::var_os("CARGO_FEATURE_STD").is_some() {
         ac.emit_expression_cfg("1f64.copysign(-1f64)", "has_copysign");
     }
+    ac.emit_expression_cfg("1u32.to_ne_bytes()", "has_int_to_from_bytes");
+    ac.emit_expression_cfg("3.14f64.to_ne_bytes()", "has_float_to_from_bytes");
 
     autocfg::rerun_path("build.rs");
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,6 +35,7 @@ pub use crate::float::FloatConst;
 pub use crate::cast::{cast, AsPrimitive, FromPrimitive, NumCast, ToPrimitive};
 pub use crate::identities::{one, zero, One, Zero};
 pub use crate::int::PrimInt;
+pub use crate::ops::bytes::ToFromBytes;
 pub use crate::ops::checked::{
     CheckedAdd, CheckedDiv, CheckedMul, CheckedNeg, CheckedRem, CheckedShl, CheckedShr, CheckedSub,
 };
--- a/src/ops/bytes.rs
+++ b/src/ops/bytes.rs
@@ -1,0 +1,384 @@
+use core::borrow::{Borrow, BorrowMut};
+use core::cmp::{Eq, Ord, PartialEq, PartialOrd};
+use core::fmt::Debug;
+use core::hash::Hash;
+use core::mem::transmute;
+
+pub trait ToFromBytes {
+    type Bytes: Debug
+        + AsRef<[u8]>
+        + AsMut<[u8]>
+        + PartialEq
+        + Eq
+        + PartialOrd
+        + Ord
+        + Hash
+        + Borrow<[u8]>
+        + BorrowMut<[u8]>
+        + Default;
+
+    /// Return the memory representation of this number as a byte array in big-endian byte order.
+    ///
+    /// # Examples
+    ///
+    /// ```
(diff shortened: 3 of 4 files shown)


Cargo.toml
```diff
@@ -18,7 +18,11 @@ features = ["std"]
...
 [dependencies]
 
+[build-dependencies]
+rustc_version = "0.2"
+
 [features]
 default = ["std"]
 std = []
 i128 = []
+int_to_from_bytes = []
```
cuviper: I don't think we need a crate feature for this. The i128 feature is sort of vestigial, as I added that before I figured out how to automatically probe for it.

build.rs
```diff
@@ -8,6 +10,19 @@ fn main() {
     } else if env::var_os("CARGO_FEATURE_I128").is_some() {
         panic!("i128 support was not detected!");
     }
+
+    match rustc_version::version_meta() {
+        Ok(ref meta) if meta.semver.major >= 1 && meta.semver.minor >= 32 => {
+            println!("cargo:rustc-cfg=int_to_from_bytes");
```
cuviper: I would rather use another `probe` here, instead of a version check.

I've also been meaning to move num's build-script probes to use the `autocfg` crate (which I also wrote), but you don't have to change that unless you're interested.

src/int.rs
```diff
@@ -278,10 +285,102 @@ pub trait PrimInt:
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use num_traits::PrimInt;
+    ///
+    /// let bytes = 0x12345678u32.to_be_bytes();
+    /// assert_eq!(bytes, [0x12, 0x34, 0x56, 0x78]);
+    /// ```
+    fn to_be_bytes(self) -> Self::Bytes;
```
cuviper: It's a breaking change to add non-defaulted methods to an existing trait, and there's no reasonable default we could provide here.

Instead, I suggest adding an new trait for this new functionality.

Cargo.toml
flier: ok, I agree

build.rs
flier: Interesting, maybe I can submit another PR to `autocfg` for method detection later.

src/int.rs
flier: agree, I added `IntToFromBytes` to `src/ops/bytes.rs`

erikbrinkman:
The PR seems to have gone fallow, but if you come back to it, it'd be great to extend this to the float types as they also got these methods in 1.40.

flier:
Hi @erikbrinkman and @cuviper

The PR has be updated 
* The trait has been renamed to `ToFromBytes`
* Use `autocfg` to check the expression
* Implements `ToFromBytes` for integer and float

src/float.rs
```diff
@@ -5,15 +5,15 @@ use core::ops::{Add, Div, Neg};
...
+use {Num, NumCast, ToFromBytes, ToPrimitive};
 
 #[cfg(all(not(feature = "std"), feature = "libm"))]
 use libm;
 
 /// Generic trait for floating point numbers that works with `no_std`.
 ///
 /// This trait implements a subset of the `Float` trait.
-pub trait FloatCore: Num + NumCast + Neg<Output = Self> + PartialOrd + Copy {
+pub trait FloatCore: Num + NumCast + Neg<Output = Self> + PartialOrd + Copy + ToFromBytes {
```
cuviper, line 16: It's a breaking change to add requirements to the existing traits. `ToFromBytes` will have to stand alone.

src/int.rs
```diff
@@ -51,6 +51,7 @@ pub trait PrimInt:
     + CheckedMul<Output = Self>
     + CheckedDiv<Output = Self>
     + Saturating
+    + ToFromBytes
```
cuviper, line 54: Also a breaking change.

cuviper (commented):
With `Bytes` implemented so generically, this could also support types like `BigUint` using `Vec<u8>`. Consuming `self` is not desirable in that case, so maybe the trait should use `&self` in the `to_` methods? And likewise for `from_`, should they take `&Self::Bytes`? I expect the references would also get optimized away for the primitive cases anyway.

We do need to drop the breaking changes to existing traits. I would also like to see unit tests that cover each of the implemented primitive types.

cuviper:
Sorry, I pushed to this branch on accident, but the revived development is in #224.

PR #104, opened 2019-03-05
Add inplace methods to `Zero` and `One`

lcnr:
Adds the following default implemented methods to `Zero` and `One`:

```rust
fn set_zero(&mut self) {
    *self = Zero::zero();
}
```

```rust
fn set_one(&mut self) {
    *self = One::one();
}
```

This allows for reuse of BigNums.

### DIFF
--- a/src/identities.rs
+++ b/src/identities.rs
@@ -1,6 +1,5 @@
 use core::num::Wrapping;
 use core::ops::{Add, Mul};
-use core::mem;
 
 /// Defines an additive identity element for `Self`.
 pub trait Zero: Sized + Add<Self, Output = Self> {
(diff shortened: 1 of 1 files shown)


lcnr:
@cuviper I looked at the guidelines and think that `set` or `make` should both be fine even if they sound somewhat odd. Changed it to `set`?  :thinking:

lcnr:
I also changed the functions to not return `&mut Self`. I now believe doing so is  a somewhat bad practice as it adds uncertainty about what exactly this method is doing.

src/identities.rs
```diff
@@ -20,6 +19,12 @@ pub trait Zero: Sized + Add<Self, Output = Self> {
     // This cannot be an associated constant, because of bignums.
     fn zero() -> Self;
 
+    /// Sets `self` to the additive identity element of `Self`, `0`.
+    /// Returns `&mut self` to enable method chaining.
```
cuviper:
```suggestion
```

PR #105, opened 2019-03-06
fix unsigned_shr docs

lcnr:


src/int.rs
```diff
@@ -171,8 +171,8 @@ pub trait PrimInt:
     /// ```
     /// use num_traits::PrimInt;
     ///
-    /// let n = 0xFEDCBA9876543210i64;
-    /// let m = 0x000FEDCBA9876543i64;
+    /// let n = 0x0123456789ABCDEFi64;
+    /// let m = 0x0000123456789ABCi64;
```
cuviper: These new values fail to demonstrate how `unsigned_shr` is different than a regular `shr`, which is that the sign bit is not preserved. This is also called a [logical shift](https://en.wikipedia.org/wiki/Logical_shift), versus an [arithmetic shift](https://en.wikipedia.org/wiki/Arithmetic_shift). The starting value needs to be negative for this to matter.

I suggest writing these literals as a decimal value, where the first is negative and the result will be positive. We can write their hex representations in comments to clarify what `unsigned_shr` is doing.

cuviper:
Thanks!

bors r+

PR #108, opened 2019-03-15
int: document PrimInt trait

### LINKED ISSUE #11: Document PrimInt
_From @gnzlbg on December 17, 2015 12:10_

It took me a bit to realize that PrimInt is the trait that provides the bit manipulation operations on primitive types (nothing more, nothing less). The methods are documented but the Trait itself is not documented.

_Copied from original issue: rust-num/num#140_

dvdhrm:
This documents the PrimInt trait, explains its intentions and features
as well as mentions its origins for future reference.

Closes #11.

### DIFF
--- a/src/int.rs
+++ b/src/int.rs
@@ -5,6 +5,32 @@ use ops::checked::*;
 use ops::saturating::Saturating;
 use {Num, NumCast};
 
+/// Generic trait for primitive integers.
+///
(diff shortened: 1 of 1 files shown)


dvdhrm:
ref #11

cuviper:
Thanks! I'm marking this to close #11 too.

bors r+

PR #118, opened 2019-06-18
Upgrating to 2018 edition

adamAndMath:
Upgrading to rust 2018 edition. This should be a non-breaking change.


cuviper:
We choose to continue during older compilers in the crate, which means we can't use the new edition. The editions are all interoperable though, so there's no functional loss from holding back.

Thanks anyway!

PR #122, opened 2019-08-17
NAN preserving clamp_lower/upper

termoshtt:
`NAN` preserving lower- and upper-clamp.

Cc: https://github.com/rust-ndarray/ndarray/issues/470#issuecomment-521809782

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -418,8 +418,8 @@ fn clamp_test() {
     assert_eq!(-1.0, clamp_lower(-2.0, -1.0));
     assert_eq!(-1.0, clamp_upper(1.0, -1.0));
     assert_eq!(-2.0, clamp_upper(-2.0, -1.0));
-    assert!(clamp_lower(std::f32::NAN, 1.0).is_nan());
-    assert!(clamp_upper(std::f32::NAN, 1.0).is_nan());
+    assert!(clamp_lower(::core::f32::NAN, 1.0).is_nan());
+    assert!(clamp_upper(::core::f32::NAN, 1.0).is_nan());
 }
 
 #[test]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -375,7 +375,7 @@ pub fn clamp<T: PartialOrd>(input: T, min: T, max: T) -> T {
 ///
 ///  If input is less than min then this returns min.
 ///  Otherwise this returns input.
-///  Preserves `NaN` different from `min`.
+///  `clamp_min(std::f32::NAN, 1.0)` preserves `NAN` different from `f32::min(std::f32::NAN, 1.0)`.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -371,17 +371,55 @@ pub fn clamp<T: PartialOrd>(input: T, min: T, max: T) -> T {
...
     // Float test
     assert_eq!(1.0, clamp(1.0, -1.0, 2.0));
     assert_eq!(-1.0, clamp(-2.0, -1.0, 2.0));
     assert_eq!(2.0, clamp(3.0, -1.0, 2.0));
+    assert_eq!(1.0, clamp_lower(1.0, -1.0));
+    assert_eq!(-1.0, clamp_lower(-2.0, -1.0));
+    assert_eq!(-1.0, clamp_upper(1.0, -1.0));
+    assert_eq!(-2.0, clamp_upper(-2.0, -1.0));
+    assert!(clamp_lower(::core::f32::NAN, 1.0).is_nan());
+    assert!(clamp_upper(::core::f32::NAN, 1.0).is_nan());
```
cuviper: Could you also add a similar `clamp` test? I think it should already be preserving NaN too.

src/lib.rs
```diff
@@ -371,17 +371,55 @@ pub fn clamp<T: PartialOrd>(input: T, min: T, max: T) -> T {
     }
 }
 
+/// A value bounded by a minimum value
+///
+///  If input is less than min then this returns min.
+///  Otherwise this returns input.
+///  Preserves `NaN` different from `min`.
```
cuviper: I have trouble understanding "different from" here. Perhaps you mean "Preserves `NaN` regardless of `min`"?

termoshtt: Fixed to: `clamp_min(std::f32::NAN, 1.0)` preserves `NAN` different from `f32::min(std::f32::NAN, 1.0)`

src/lib.rs
termoshtt: Added

src/lib.rs
cuviper: Ah, I thought you were referring to the `min` parameter. It makes more sense now.

src/lib.rs
```diff
@@ -371,17 +371,56 @@ pub fn clamp<T: PartialOrd>(input: T, min: T, max: T) -> T {
...
+    if input < min {
+        min
+    } else {
+        input
+    }
+}
+
+/// A value bounded by a maximum value
+///
+///  If input is less than min then this returns min.
```
cuviper:
```suggestion
///  If input is greater than max then this returns max.
```

cuviper:
Thanks!

bors r+

Issue #125, opened 2019-08-30
`from_str_radix` panics when the first char is encoded with multiple bytes


HeroicKatora:
[play](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9162a4a41145c8de6ad8b741c2a1fe5c).

```rust
fn main() {
    let _ = f32::from_str_radix("™0.2", 10);
}
```

## Expected behaviour

Nothing hapens. The method returns an error with kind `Invalid` which is ignored.

## Observed behaviour

The code panics.

```
thread 'main' panicked at 'byte index 1 is not a char boundary; it is inside '™' (bytes 0..3) of `™0.2`', src/libcore/str/mod.rs:2034:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

## Cause analysis

The underlying cause is that `slice_shift_char` splits the string at a hardcode byte offset of `1` instead of the byte-length of the first character. This panics when it is not a UTF-8 byte boundary.

https://github.com/rust-num/num-traits/blob/58f02a8677c13ad5d46d0eb0698d9052ca70ef73/src/lib.rs#L218-L220

HeroicKatora:
Possible fixed implementation:

```
fn slice_shift_char(src: &str) -> Option<(char, &str)> {
     let mut chars = src.chars();
     if let Some(ch) = chars.next() {
         Some((ch, chars.as_str())) // available since Rust 1.4.0
     } else {
         None
     }
 }
```

cuviper:
Thanks for the report, and your suggested fix looks great! Can you submit this as a pull request along with corresponding tests? It looks like `slice_shift_char` is used both at the beginning of the string and in the exponent position (for input like `1e100`).

PR #126, opened 2019-08-30
Fix num parsing for invalid multi-byte sign chars

HeroicKatora:
Ensure that splitting the potential sign character from the remainder
respects UTF8 boundaries. This lets invalid characters fail correctly
with an error, instead of panicking.

Closes: #125


cuviper:
Thanks!

bors r+

Issue #121, opened 2019-08-09
Add traits for generic norms and distances.


Nuramon27:
In many iterative numerical algorithms, one has to calculate the distance between a solution candidate and a reference value in order to check for convergence. But currently there is no trait which would enable the computation of such a distance generically. 

There is `Signed` which provides an `abs` function, but this trait can only be implemented for types with a notion of signedness. The correct mathematical concept, however, are norms and metrics, with the latter being reasonable for basically _any_ numeric type.

So I propose adding two traits, `Norm` and `Distance`, each with one method of the same name. The `norm`-function of the `Norm` trait returns the norm of a _single_ value, the `distance`-function of the `Distance` trait returns the distance between _two_ values.

In my branch [dist](https://github.com/Nuramon27/num-traits/tree/dist) I have written a draft of these traits and implemented them for the primitive numeric types.

cuviper:
It's a reasonable idea. Can you open it as a PR for discussion?

One thing to consider is that there are multiple possible definitions of norm, especially when more dimensions are added like in `Complex<T>`. e.g. [Euclidean norm](https://en.wikipedia.org/wiki/Norm_(mathematics)#Euclidean_norm) vs. [Manhattan norm](https://en.wikipedia.org/wiki/Norm_(mathematics)#Taxicab_norm_or_Manhattan_norm).

JohnScience:
> It's a reasonable idea. Can you open it as a PR for discussion?
> 
> One thing to consider is that there are multiple possible definitions of norm, especially when more dimensions are added like in `Complex<T>`. e.g. [Euclidean norm](https://en.wikipedia.org/wiki/Norm_(mathematics)#Euclidean_norm) vs. [Manhattan norm](https://en.wikipedia.org/wiki/Norm_(mathematics)#Taxicab_norm_or_Manhattan_norm).

Are you sure that it must be the responsibility of num-traits?
Creating another managed crate (for example, "num-maths-traits") which would contain category-, group-, field-, ring-, space-, and other more general traits sounds a better idea. In this case, trait inheritance from more general traits can be an opt-in feature enabled by default.

The problem is not even that num-traits may end up becoming too of a heavy dependency in attempts to fit all needs but that num-traits' purpose statement ("Numeric traits for generic mathematics in Rust"), would, in this case, become unclear to me and, perhaps, someone else.

cuviper:
> Are you sure that it must be the responsibility of num-traits?

[...]

PR #130, opened 2019-09-01
Add traits for generic norms and distances

Nuramon27:
This is the implementation mentioned in #121.
Regarding the issue that there are many possible norms for a certain type (see comment https://github.com/rust-num/num-traits/issues/121#issuecomment-526777142 ): I think these traits would only be intended to be implemented on very basic types for which there is some notion of a _standard norm_, like the absolute value on floating point numbers and the euclidean norm on complex numbers and one-dimensional arrays.

Indeed in the meantime I started working a separate crate [Norman](https://gitlab.com/Nuramon27/norman) which handles exactly this issue and provides a set of different norms on a single type.

I hoped that both might go well with each other: having num-traits for a simple answer to the question “how far are these two vectors apart” and a dedicated crate if full control is necessary over how this distance is calculated.

### DIFF
--- a/src/dist.rs
+++ b/src/dist.rs
@@ -1,21 +1,75 @@
+//! Traits for generically calculating distances between values.
+//!
+//! This is often necessary in generic numeric algorithms to determine if
+//! the demanded precision is reached.
+
 use std::ops::{Sub, Div, DivAssign};
 
 use {Num};
 
+/// The abstract notion of the distance between two values.
+///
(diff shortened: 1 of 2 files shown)


anvil777:
You could check out my [`maths-traits`](https://github.com/anvil777/maths-traits) crate. While it doesn't _currently_ have `impl`s for the structs in the `num` collection, there are norm, metric, and inner product traits, and I'd be open to contributions or feature requests.

PR #132, opened 2019-09-08
Add comments explaining why transmutes are safe

Shnatsel:
Add comments explaining why transmutes are safe so that people auditing unsafe code don't have to spend time figuring it out by themselves.


cuviper:
Thanks!

bors r+

PR #135, opened 2019-09-24
Debug-panic in clamp_min/max if min/max is NAN

### LINKED ISSUE #134: Behavior of NAN as min/max for clamp_min/max
#122 added `clamp_min` and `clamp_max` functions which preserve NANs in the `input`. I didn't see any discussion about what the best behavior is when the `min`/`max` argument is NAN, and I think it's worth considering before `clamp_min`/`clamp_max` are included in an official release. I think four different possible behaviors are reasonable when `min`/`max` is NAN:

1. Panic in debug mode
2. Panic in both debug and release mode
3. Return `input`
4. Return NAN

Option 1 matches `num_traits::clamp`. Option 2 matches (currently unstable) `f32::clamp`/`f64::clamp`. Option 3 is what the implementation currently does.

I lean towards option 1 for consistency with `num_traits::clamp` and because it avoids the question of whether option 3 or option 4 makes more sense.

Thoughts?

cc @termoshtt @cuviper

jturner314:
This also improves the docs for `clamp`, `clamp_min`, and `clamp_max`.

Closes #134.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -437,27 +437,38 @@ fn clamp_test() {
 }
 
 #[test]
-fn clamp_nan_bound() {
-    /// When debug assertions and the `std` feature are enabled, checks that
-    /// the expression panics.
-    macro_rules! assert_debug_panics {
-        ($body:expr) => {
-            #[cfg(all(debug_assertions, feature = "std"))]
-            {
-                if let Ok(v) = ::std::panic::catch_unwind(|| $body) {
-                    panic!(
-                        "assertion failed: should_panic; non-panicking result: {:?}",
-                        v
-                    );
-                }
-            }
-        };
-    }
-    assert_debug_panics!(clamp(0., ::core::f32::NAN, 1.));
-    assert_debug_panics!(clamp(0., -1., ::core::f32::NAN));
-    assert_debug_panics!(clamp(0., ::core::f32::NAN, ::core::f32::NAN));
-    assert_debug_panics!(clamp_min(0., ::core::f32::NAN));
-    assert_debug_panics!(clamp_max(0., ::core::f32::NAN));
+#[should_panic]
+#[cfg(debug_assertions)]
+fn clamp_nan_min() {
+    clamp(0., ::core::f32::NAN, 1.);
+}
+
+#[test]
+#[should_panic]
+#[cfg(debug_assertions)]
+fn clamp_nan_max() {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -428,6 +436,28 @@ fn clamp_test() {
     assert!(clamp_max(::core::f32::NAN, 1.0).is_nan());
 }
 
+#[test]
+fn clamp_nan_bound() {
+    /// When debug assertions are enabled, checks that the expression panics.
+    macro_rules! assert_debug_panics {
+        ($body:expr) => {
+            if cfg!(debug_assertions) {
+                if let Ok(v) = ::std::panic::catch_unwind(|| $body) {
```
cuviper: I think you'll need to guard this test for the `std` feature, and you might as well check debug in the same guard -- `#[cfg(all(debug_assertions, feature = "std"))]`. Or you could split them into separate `#[should_panic]` tests, and not need `std`.

Note that using `cfg!` only creates a `bool`, but everything within still has to resolve.

jturner314: Oh, good catch. I've updated the PR.

jturner314: Would you prefer `#[should_panic]` tests? That would look like this:

```rust
#[test]
#[should_panic]
#[cfg(debug_assertions)]
fn clamp_nan_min() {
    clamp(0., ::core::f32::NAN, 1.);
}

#[test]
#[should_panic]
#[cfg(debug_assertions)]
fn clamp_nan_max() {
    clamp(0., -1., ::core::f32::NAN);
}

#[test]
#[should_panic]
#[cfg(debug_assertions)]
fn clamp_nan_min_max() {
    clamp(0., ::core::f32::NAN, ::core::f32::NAN);
}

#[test]
#[should_panic]
#[cfg(debug_assertions)]
fn clamp_min_nan_min() {
    clamp_min(0., ::core::f32::NAN);
}

#[test]
#[should_panic]
#[cfg(debug_assertions)]
fn clamp_max_nan_max() {
    clamp_max(0., ::core::f32::NAN);
}
```

It's longer, but the additional clarity and ability to run in `no_std` contexts is nice.

jturner314: From the CI results, it looks like using `should_panic` is the cleanest way to support Rust 1.8 (which allows attributes only on items). I've updated the PR.

cuviper:
Thanks!

bors r+

PR #137, opened 2019-10-18
Document NumCast / ToPrimitive's sucess vs error rules

### LINKED ISSUE #9: Document NumCast / ToPrimitive's sucess vs error rules
_From @bluss on March 7, 2016 4:43_

The traits `NumCast` and `ToPrimitive` use `Option<T>` return values, they need to document when and why they can return `None`.

_Copied from original issue: rust-num/num#177_

ocstl:
Closes #9 .

The added comments are pretty succinct, not going into details of the conversion (for example, truncation when converting from float to int). If something is missing, I'm not adverse to adding precision.

### DIFF
--- a/src/cast.rs
+++ b/src/cast.rs
@@ -10,38 +10,45 @@ use float::FloatCore;
 
 /// A generic trait for converting a value to a number.
 pub trait ToPrimitive {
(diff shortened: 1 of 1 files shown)


cuviper:
Thanks!

bors r+

PR #144, opened 2019-11-21
Update to libm 0.2.0

CryZe:
libm 0.2.0 was just released with a bunch of bugfixes. Since none of the public API of libm is exposed, this should be a non-breaking change. Unfortunately they removed the extension traits, so this implementation isn't quite as neat anymore as before.


src/float.rs
```diff
@@ -1889,7 +1889,7 @@ macro_rules! float_impl_std {
 
 #[cfg(all(not(feature = "std"), feature = "libm"))]
 macro_rules! float_impl_libm {
-    ($T:ident $decode:ident $LibmImpl:ident) => {
+    ($T:ident $decode:ident { $($LibmImpl:tt)* }) => {
```
cuviper: I think rather than `tt`-munching, it would be nicer to now flip the macro around and just fill in the common methods. Something like:

```rust
impl Float for f32 {
    float_impl_libm!(...) // common stuff
    // then f32-specific items
}
impl Float for f64 {
    float_impl_libm!(...) // common stuff
    // then f64-specific items
}

CryZe: Alright, I inverted it.

cuviper:
Thanks -- I went ahead and bumped the version for release.

bors r+

PR #145, opened 2019-11-26
Add the full circle constant τ (TAU)

m-ou-se:
Since it is not (yet?) added to the standard library, the value is included in the source here, instead of referring to a constant in `core`.

### DIFF
--- a/src/float.rs
+++ b/src/float.rs
@@ -2247,6 +2247,7 @@ macro_rules! float_const_impl {
         #[allow(non_snake_case)]
         pub trait FloatConst {
             $(#[$doc] fn $constant() -> Self;)+
+            #[doc = "Return the full circle constant `τ`."] fn TAU() -> Self;
         }
         float_const_impl! { @float f32, $($constant,)+ }
         float_const_impl! { @float f64, $($constant,)+ }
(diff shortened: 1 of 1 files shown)


src/float.rs
```diff
@@ -2247,6 +2247,7 @@ macro_rules! float_const_impl {
         #[allow(non_snake_case)]
         pub trait FloatConst {
             $(#[$doc] fn $constant() -> Self;)+
+            #[doc = "Return the full circle constant `τ`."] fn TAU() -> Self;
```
cuviper: It's a breaking change to add items to a trait without a default implementation, and there are no super-traits on `FloatConst` to give us the means to compute a default. However, if you add a condition `where Self: Add<Self, Output = Self>` on this method, we can add a default `Self::PI() + Self::PI()`.

m-ou-se: Sure. Done.

cuviper:
Thanks!

bors r+

stijnfrishert:
I'd really like to make use of this. Is there a reason the latest version on crates.io is 2.1?

cuviper:
It was published in `num-traits 0.2.11`, and `num 0.2.1` also requires that.

PR #147, opened 2020-01-02
Add documentation of compile-time width function alternative.

dcsommer:
While the `PrimInt` trait itself doesn't have compile time width functions, there are other ways to find the width of the underlying type. It is useful to present the alternative.

### DIFF
--- a/src/int.rs
+++ b/src/int.rs
@@ -22,7 +22,7 @@ use {Num, NumCast};
 ///
 /// All `PrimInt` types are expected to be fixed-width binary integers. The width can be queried
 /// via `T::zero().count_zeros()`. The trait currently lacks a way to query the width at
-/// compile-time.
+/// compile-time, but `core::mem::size_of::<T>()` can be used in the meantime.
 ///
 /// While a default implementation for all builtin primitive integers is provided, the trait is in
 /// no way restricted to these. Other integer types that fulfil the requirements are free to


src/int.rs
```diff
@@ -22,7 +22,7 @@ use {Num, NumCast};
 ///
 /// All `PrimInt` types are expected to be fixed-width binary integers. The width can be queried
 /// via `T::zero().count_zeros()`. The trait currently lacks a way to query the width at
-/// compile-time.
+/// compile-time, but `core::mem::size_of::<T>()` can be used in the meantime.
```
cuviper: Should we say `8 * size_of()` so it counts bits?

dcsommer:
Yeah, I can qualify counting bytes vs bits in the text. Typically the size
of integers is specified by bytes, not bits, however. Do you think all
instances of the pattern `T::zero().count_zeros()` should be rewritten in
terms of `core::mem::size_of::<T>()` instead? I could modify the text to
that effect if you like.

cuviper:
> Typically the size of integers is specified by bytes, not bits, however.

You think so? The bit width is right in the name: `i32`, `u8`, etc., and C's `stdint.h` is similar.

> Do you think all instances of the pattern `T::zero().count_zeros()` should be rewritten in terms of `core::mem::size_of::<T>()` instead?

AFAICS "all" is just this one instance, but sure, I'd be OK with preferring `size_of` since it's `const`.

dcsommer:
I just remembered that in the case of `dyn T` we can't use
`core::mem::size_of<T>`, so there is definitely a use case for both runtime
and compile-time width calculation. I'll do it in terms of bits too.

cuviper:
How does `dyn T` apply in this context though? We already require `PrimInt: Sized`.

dcsommer:
I mean when dealing with a `PrimInt` trait object, we can't know the size
of the type at compile time. For instance, in code dealing with a `Box<dyn
PrimInt>`, there isn't a way to use the `const fn`
`core::mem::size_of<T>()` to get the concrete type's width at compile time.
We need to dynamic dispatch on the trait object to calculate it.

cuviper:
`PrimInt` is not trait object safe -- you can't form any sort of `dyn PrimInt` at all.

Issue #95, opened 2019-01-08
PrimInt lacks the wrapping ops and overflowing ops


dcrewi:
Is this a design decision or has nobody needed it yet?

The wrapping ops are available as via the `ops::wrapping::*` traits, but it seems like they ought to be a part of PrimInt too. The overflowing ops are completely missing, as far as I can see.

dcrewi:
Okay.

But what about another set of traits for the overflowing ops?

cuviper:
New traits for overflowing ops would be fine with me.

PR #153, opened 2020-02-05
Add WrappingNeg trait

### LINKED ISSUE #10: Wrapping neg is not implemented for PrimInt
_From @gnzlbg on December 23, 2015 18:40_

When constraining on `PrimInt` there is no way of performing wrapping neg in a generic way.

The "generic" way to do this seems to be to call the `.wrapping_neg()` method which is offered by all primitive integers (but not by `PrimInt`).

_Copied from original issue: rust-num/num#146_

ocstl:
Closes #10 

A quick search turned up that the [extprim](https://github.com/kennytm/extprim) implements `PrimInt` for its `u128` and `i128`, but, since both have their own `wrapping_neg` method, it's a non-issue.

That being said, I don't know for certain that it is the only project that implements `PrimInt`, so it might be wise to wait for a larger version bump before merging this.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -43,7 +43,9 @@ pub use ops::checked::{
 pub use ops::inv::Inv;
 pub use ops::mul_add::{MulAdd, MulAddAssign};
 pub use ops::saturating::Saturating;
-pub use ops::wrapping::{WrappingAdd, WrappingMul, WrappingShl, WrappingShr, WrappingSub};
+pub use ops::wrapping::{
+    WrappingAdd, WrappingMul, WrappingNeg, WrappingShl, WrappingShr, WrappingSub,
+};
 pub use pow::{checked_pow, pow, Pow};
 pub use sign::{abs, abs_sub, signum, Signed, Unsigned};
 
--- a/src/ops/wrapping.rs
+++ b/src/ops/wrapping.rs
@@ -116,7 +116,8 @@ pub trait WrappingNeg: Sized {
     /// use num_traits::WrappingNeg;
     ///
     /// assert_eq!(100i8.wrapping_neg(), -100);
-    /// assert_eq!((-128i8).wrapping_neg(), -128);
+    /// assert_eq!((-100i8).wrapping_neg(), 100);
+    /// assert_eq!((-128i8).wrapping_neg(), -128); // wrapped!
     /// ```
     fn wrapping_neg(&self) -> Self;
 }
--- a/src/ops/wrapping.rs
+++ b/src/ops/wrapping.rs
@@ -1,5 +1,5 @@
 use core::num::Wrapping;
-use core::ops::{Add, Mul, Shl, Shr, Sub};
+use core::ops::{Add, Mul, Neg, Shl, Shr, Sub};
 
 macro_rules! wrapping_impl {
     ($trait_name:ident, $method:ident, $t:ty) => {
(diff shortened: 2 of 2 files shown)


cuviper:
Thinking about #10 again -- we don't have any other "wrapping" methods on `PrimInt`, so I'm not sure why we would add `wrapping_neg` in particular. See also #95.

However, we do have `WrappingAdd` etc. -- so how about adding `WrappingNeg`?

cuviper:
> A quick search turned up that the [extprim](https://github.com/kennytm/extprim) implements `PrimInt` for its `u128` and `i128`, but, since both have their own `wrapping_neg` method, it's a non-issue.

Note that while they do have an inherent `wrapping_neg` method, they would still be broken by a new `PrimInt` with that method until they add it to their [implementation](https://github.com/kennytm/extprim/blob/7bd81c12787eec04405dbea63c47cf2928339645/src/u128.rs#L1914) explicitly.
(Rust doesn't use structural typing for traits, the way Go does for interfaces.)

ocstl:
> Note that while they do have an inherent wrapping_neg method, they would still be broken by a new PrimInt with that method until they add it to their implementation explicitly.
> (Rust doesn't use structural typing for traits, the way Go does for interfaces.)

I thought I had tested for it, but you made me realise where I made a mistake. Thank you!

> Thinking about #10 again -- we don't have any other "wrapping" methods on PrimInt, so I'm not sure why we would add wrapping_neg in particular. See also #95.
> 
> However, we do have WrappingAdd etc. -- so how about adding WrappingNeg?

It did feel a bit odd. And thanks for pointing out #95, it puts things in context.

The potential issue with adding WrappingNeg is that Neg was only implemented for Wrapping in Rust 1.10. Wouldn't this create a problem with regards to maintaining compatibility with 1.8?

cuviper:
> The potential issue with adding WrappingNeg is that Neg was only implemented for Wrapping in Rust 1.10. Wouldn't this create a problem with regards to maintaining compatibility with 1.8?

I think it will be OK if we follow the style of the others:

```rust
impl<T: WrappingNeg> WrappingNeg for Wrapping<T>
where
    Wrapping<T>: Neg<Output = Wrapping<T>>,
{
    fn wrapping_neg(&self) -> Self {
        Wrapping(self.0.wrapping_neg())
    }
}
```

That `where` constraint just won't match anything before Rust 1.10. As written, it's not actually _using_ `<Wrapping<T> as Neg>::neg` at all, so this is overconstrained, but it's what we have on the others. :man_shrugging:

ocstl:
One thing about this implementation that bugs me a bit is that the `WrappingNeg` trait can't be bound on `Neg`, since that trait is not implemented for the unsigned types, for (in retrospect) obvious reasons. This breaks with the style for the other Wrapping traits.

> That where constraint just won't match anything before Rust 1.10. As written, it's not actually using <Wrapping<T> as Neg>::neg at all, so this is overconstrained, but it's what we have on the others.

That is good to know, thank you!

I didn't see it coming, but turns out the absence of the Neg trait for Wrapping in Rust 1.8 is causing issues for testing though. I guess a workaround would be to use Sub instead (`Wrapping(0_u8) - Wrapping(...)`), but this seems unsatisfactory.

While it would be great to be able to close this issue, if you feel any discomfort at all about this, don't hesitate to say so.

ocstl:
> One thing about this implementation that bugs me a bit is that the WrappingNeg trait can't be bound on Neg, since that trait is not implemented for the unsigned types, for (in retrospect) obvious reasons. This breaks with the style for the other Wrapping traits.

Since CheckedNeg isn't bound by Neg either, this is fine. I tried to maintain the same order of operations as the Checked traits to make comparisons easier.

I also fixed a slight formatting issue.

The test for Wrapping falls back on Sub for the time being for lack of a better idea, although it does work for versions including 1.10 and above; I added a FIXME in case Neg becomes available following a MSRV change.

Thank you for your time, and let me know if there's anything.

ocstl:
Sorry for the stream-of-thought like nature of the previous comments.

> That where constraint just won't match anything before Rust 1.10. As written, it's not actually using <Wrapping<T> as Neg>::neg at all, so this is overconstrained, but it's what we have on the others. 

This creates a bit of a problem for the `wrapping_is_wrappingneg()` test function, since in Rust 1.8, Wrapping is not WrappingNeg, although it is once we reach 1.10. This bothered me a bit, but I guess that since Neg is not implemented for Wrapping over all covered versions, this is not something one can guarantee to users.

Also, Since Neg was only implemented in 1.10, we can't actually test the Neg op (`-Wrapping(255u8)`). I replaced both of these tests with a TODO in a separate commit. This way, if the MSRV reaches 1.10, it's a simple matter of reverting it to bring back the tests.

Alternatively, if you'd rather tabled implemented WrappingNeg if and when the MSRV reaches 1.10 so both tests can be included, I would understand completely.

This legacy business is tricky. So, thank you for your hard work. :bowing_man:

cuviper:
I added a little bit to the doc example, but otherwise I think this is good, thanks!

bors r+

PR #155, opened 2020-02-13
Added copysign fixes #152

### LINKED ISSUE #152: Float trait doesn't require copysign function
`Float` trait should require `copysign` function introduced in Rust 1.35: [f32::copysign](https://doc.rust-lang.org/std/primitive.f32.html#method.copysign), [f64::copysign](https://doc.rust-lang.org/std/primitive.f64.html#method.copysign)

peartt:


cuviper:
Please look at the failed CI results, and let me know if you need help...

src/float.rs
```diff
@@ -1806,6 +1806,37 @@ pub trait Float: Num + Copy + NumCast + PartialOrd + Neg<Output = Self> {
...
+    /// assert_eq!(f.copysign(-0.42), -3.5_f32);
+    /// assert_eq!((-f).copysign(0.42), 3.5_f32);
+    /// assert_eq!((-f).copysign(-0.42), -3.5_f32);
+    ///
+    /// assert!(f32::nan().copysign(1.0).is_nan());
+    /// ```
+    fn copysign(self, sign:Self) -> Self {
+        if self.is_nan() {
+            sign / Self::zero()
+        } else if self.signum() == sign.signum() {
```
tspiteri, line 1834: `is_sign_negative` is better than `signum` here as it works on NaNs as well; `signum` itself is implemented using `is_sign_negative` in the trait.

Then the special case for NaN can be removed as both `is_sign_negative` and `neg` work well on NaNs. (Currently the NaN special case is buggy anyway and returns infinity rather than NaN.)

XAMPPRocky:
@peartt Hello, are you still interested in continuing this PR? 🙂 As I would be interested in seeing this merged so we can use it in a project, and would be willing to continue the work.

PR #161, opened 2020-04-20
CI: cache cargo registry

cuviper:


cuviper:
I don't think this is saving as much as I wanted...

PR #165, opened 2020-04-30
Add saturating_mul() and refactor Saturating into subtraits. Fixes #40.

### LINKED ISSUE #40: `saturating_mul` missing
Is the lack of `saturating_mul` in `num::Saturating` on purpose, or a mistake, or should it live somewhere else?

trepetti:
Hi all,

Taking the suggestion from #40, this pull request breaks out the saturating trait into three separate SaturatingAdd, SaturatingSub and SaturatingMul traits, borrowing the idea from Checked* and Wrapping*. In terms of native ops, SaturatingNeg would be an option on signed types once the saturating_neg() API is stable on signed integer primitives.

I know there is concern about breaking changes, so I was thinking that I could add a wrapping Saturating trait for backwards compatibility that retains the original functionality:

```
pub trait Saturating: SaturatingAdd + SaturatingSub {}
impl<T> Saturating for T where T: SaturatingAdd + SaturatingSub {}
```

This is not included in the current version, but I would be happy to add.

-Tom (trepetti@cs.columbia.edu)

### DIFF
--- a/src/ops/saturating.rs
+++ b/src/ops/saturating.rs
@@ -1,5 +1,37 @@
 use core::ops::{Add, Mul, Sub};
 
+/// Saturating math operations. Deprecated, use `SaturatingAdd`, `SaturatingSub` and
+/// `SaturatingMul` instead.
+pub trait Saturating {
+    /// Saturating addition operator.
+    /// Returns a+b, saturating at the numeric bounds instead of overflowing.
+    fn saturating_add(self, v: Self) -> Self;
+
+    /// Saturating subtraction operator.
+    /// Returns a-b, saturating at the numeric bounds instead of overflowing.
+    fn saturating_sub(self, v: Self) -> Self;
+}
+
+macro_rules! deprecated_saturating_impl {
+    ($trait_name:ident for $($t:ty)*) => {$(
+        impl $trait_name for $t {
+            #[inline]
+            fn saturating_add(self, v: Self) -> Self {
+                Self::saturating_add(self, v)
+            }
+
+            #[inline]
+            fn saturating_sub(self, v: Self) -> Self {
+                Self::saturating_sub(self, v)
+            }
+        }
+    )*}
+}
+
+deprecated_saturating_impl!(Saturating for isize usize i8 u8 i16 u16 i32 u32 i64 u64);
+#[cfg(has_i128)]
+deprecated_saturating_impl!(Saturating for i128 u128);
+
 macro_rules! saturating_impl {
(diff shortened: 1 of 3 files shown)


trepetti:
Also, it occurred to me that there is no harm in making a backwards compatible compound Saturating trait from `SaturatingAdd + SaturatingSub + SaturatingMul` since there are no primitive integer types that support the first two without SaturatingMul. So that may be an option, too.

cuviper:
> I know there is concern about breaking changes, so I was thinking that I could add a wrapping Saturating trait for backwards compatibility that retains the original functionality:
> 
> ```
> pub trait Saturating: SaturatingAdd + SaturatingSub {}
> impl<T> Saturating for T where T: SaturatingAdd + SaturatingSub {}
> ```

Breaking changes are definitely a concern, and unfortunately this idea doesn't solve it. That would be fine for _users_ of the trait, but it still breaks any implementors of the trait. Since the trait is not sealed, 3rd parties can and do implement it as well, e.g. [`extprim::u128`](https://docs.rs/extprim/1.7.0/extprim/u128/struct.u128.html#impl-Saturating). That concern applies to the constraint change on `PrimInt` too.

We can instead _deprecate_ `Saturating`, nudging folks to use the new separate traits. `PrimInt` will be stuck with the old one, but that's life.

trepetti:
Good point! Sorry I did not think that through entirely. 

I restored the original `Saturating` trait and updated the doc comment to mention it has been deprecated. Similarly, I restored `PrimInt` to its original state with just `Saturating` and now expose all four `Saturating`, `SaturatingAdd`, `SaturatingSub` and `SaturatingMul` in `lib.rs`.

Let me know if there is a better way to go about deprecating `Saturating`, if that is the route you want to take. Looking back, it seemed that just mentioning as much in the doc comments is what has been done historically. 

Thank you for looking at my pull request.

src/ops/saturating.rs
```diff
@@ -25,6 +28,118 @@ macro_rules! saturating_impl {
...
+    ($trait_name:ident, $method:ident, $t:ty) => {
+        impl $trait_name for $t {
+            #[inline]
+            fn $method(&self, v: &Self) -> Self {
+                <$t>::$method(*self, *v)
+            }
+        }
+    };
+    ($trait_name:ident, $method:ident, $t:ty, $rhs:ty) => {
+        impl $trait_name<$rhs> for $t {
```
cuviper, line 45: Is this second macro pattern actually used? I think not, because your new traits don't have a generic RHS.

PR #173, opened 2020-06-18
Add license terms to README

### LINKED ISSUE #166: Clarify licensing
Hi !

This repo has both the Apache and MIT licenses, as most of the main rust repo, but is lacking a copyright notice indicating that any of the two can be chosen at the discretion of the user. Rust's main repo uses the following one: https://github.com/rust-lang/rust/blob/master/COPYRIGHT.

I am not a lawyer but I believe that without an explicit mention that the user is free to choose, the most restrictive license (which I guess is the Apache) applies.

Thanks a lot for your work

newpavlov:
Closes #166

This project already has LICENSE-* files and I don't think an additional COPYRIGHT file is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,3 +52,18 @@ Release notes are available in [RELEASES.md](RELEASES.md).
 ## Compatibility
 
 The `num-traits` crate is tested for rustc 1.8 and greater.
+
+## License
+
+Licensed under either of
+
+ * [Apache License, Version 2.0](http://www.apache.org/licenses/LICENSE-2.0)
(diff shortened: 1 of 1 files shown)


cuviper:
Hi, and thanks!

We should put this on all the num crates, if you're interested, or I'll try to remember it next time I publish them.

bors r+

newpavlov:
Right now I only use this crate, so I will leave other crates to you.

cuviper:
> so I will leave other crates to you.

That's fine, thanks!

PR #176, opened 2020-07-05
Document cast traits about possible precision loss and truncation

### LINKED ISSUE #175: Misleading ToPrimitive / NumCast behavior to a float type with precision loss
The documentation of the traits `ToPrimitive` and `NumCast` usually describe that it returns `None` when the source value cannot be represented in the target type.

([quote](https://docs.rs/num-traits/0.2.12/num_traits/cast/trait.NumCast.html#tymethod.from))

> Creates a number from another value that can be converted into a primitive via the ToPrimitive trait. If the source value cannot be represented by the target type, then None is returned.

However, in practice these methods will not return `None` when converting integer values which cannot be _fully_ represented (without precision loss) as a floating point type. As an example, `ToPrimitive` admits a conversion from `16_777_217_u32` to `f32`, losing one bit of precision in the process:

```
let x: u32 = 16_777_217;
let y = x.to_f32();
println!("{:?}", y);
```
```none
Some(16777216.0)
```
[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f73687d2e63d97c4619f375e44083f76)

I am not entirely familiar with how the current implementation handles conversions to floating point types, but it feels that what is indicated as "can or cannot be represented" is worth clarifying in the documentation.

Enet4:
Resolves #175.
Document the traits `ToPrimitive`, `FromPrimitive` and `NumCast` that
value can be represented by the target type
when they are within the range admitted by the target type,
and so truncation or precision loss may happen.

One question worth reflecting on is whether this documentation be normalized somewhere at module level,
and whether it should be replicated to all methods or simply stay at each trait.
I would also be open to adding examples with doctests if deemed useful.

### DIFF
--- a/src/cast.rs
+++ b/src/cast.rs
@@ -9,6 +9,14 @@ use core::{u16, u32, u64, u8, usize};
 use float::FloatCore;
 
 /// A generic trait for converting a value to a number.
+///
+/// A value can be represented by the target type when it lies within
+/// the range of scalars supported by the target type.
+/// For example, a negative integer cannot be represented by an unsigned
+/// integer type, and an `f64` with a very high magnitude might not be
+/// convertible to an `f32`.
+/// On the other hand, conversions with possible precision loss or truncation
+/// (e.g. an `f32` with a decimal part to an integer type) are admitted.
 pub trait ToPrimitive {
     /// Converts the value of `self` to an `isize`. If the value cannot be
     /// represented by an `isize`, then `None` is returned.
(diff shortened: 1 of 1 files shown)


cuviper:
This looks good to me, thanks!

bors r+

> One question worth reflecting on is whether this documentation be normalized somewhere at module level,

While it _could_ go on the `cast` module, these traits are also re-exported at the root, so I don't know how often folks would even see the module-level documentation. 

> and whether it should be replicated to all methods or simply stay at each trait.

I think that would be overkill. The trait-level docs are already on the same page as all the methods.

> I would also be open to adding examples with doctests if deemed useful.

It's clear enough _to me_ what is meant by precision loss and truncation, but if you'd like to follow up with some examples, that would be fine.

Issue #168, opened 2020-05-22
No Overflowing* traits.


human-0:
There are currently no `Overflowing*` traits like there are `Checked` and `Wrapping` traits.

cuviper:
You're right, but this seems like a fine idea. #165 is also adding more `Saturating` traits.

cuviper:
Closed by #180.

PR #180, opened 2020-07-17
Add Overflowing trait

KaczuH:
Closes issue #168

I've been inspired by `saturated.rs`

### DIFF
--- a/src/ops/overflowing.rs
+++ b/src/ops/overflowing.rs
@@ -15,9 +15,10 @@ macro_rules! overflowing_impl {
     };
 }
 
-/// If an overflow would have occurred then the wrapped value is returned.
+/// Performs addition with a flag for overflow.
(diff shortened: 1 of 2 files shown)


src/ops/overflowing.rs
```diff
@@ -0,0 +1,101 @@
...
+        impl $trait_name for $t {
+            #[inline]
+            fn $method(&self, v: &Self) -> (Self, bool) {
+                <$t>::$method(*self, *v)
+            }
+        }
+    };
+}
+
+/// If an overflow would have occurred then the wrapped value is returned.
```
cuviper, line 18: This line seems like it would better belong as the second sentence of the method docs. At the top we need a more general description, perhaps something simple like "Performs addition with a flag for overflow."

The same goes for OverflowingSub/Mul.

cuviper:
LGTM, thanks!

bors r+

PR #184, opened 2020-07-30
Update AsPrimitive Safety docs for rust 1.45

### LINKED ISSUE #174: Warning in docs of num_traits::cast::AsPrimitive seems to be out-of-date
The documentation of [`num_traits::cast::AsPrimitive`](https://docs.rs/num-traits/0.2.12/num_traits/cast/trait.AsPrimitive.html) warns about undefined behavior:

> Currently, some uses of the as operator are not entirely safe. In particular, it is undefined behavior if:
> ...
> Or a floating point value does not fit in another floating point type (#15536)

However, [issue 15536](https://github.com/rust-lang/rust/issues/15536) is closed and fixed, as far as I understand, and this is not undefined behavior anymore (in fact, it was never really undefined behavior; the LLVM docs described it as such, but those docs have been clarified).

It looks like the warning in the documentation of `AsPrimitive` is not accurate anymore.

martin-t:
Fixes #174

The safety section should not be removed entirely since people might be using older compilers (as was pointed out to me in a [related issue](https://github.com/yoanlcq/vek/issues/59) in the vek crate).

However, [this](https://github.com/rust-lang/rust/issues/15536#issuecomment-377681267) and followup comments indicate that the second case (float to float) was never UB in the first place - should the second example be removed?

### DIFF
--- a/src/cast.rs
+++ b/src/cast.rs
@@ -752,8 +752,8 @@ impl<T: NumCast> NumCast for Wrapping<T> {
 ///
 /// # Safety
 ///
-/// Currently, some uses of the `as` operator are not entirely safe.
-/// In particular, it is undefined behavior if:
+/// **In Rust versions before 1.45.0**, some uses of the `as` operator were not entirely safe.
+/// In particular, it was undefined behavior if:
 ///
 /// - A truncated floating point value cannot fit in the target integer
 ///   type ([#10184](https://github.com/rust-lang/rust/issues/10184));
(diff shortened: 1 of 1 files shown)


cuviper:
Hi,

> However, [this](https://github.com/rust-lang/rust/issues/15536#issuecomment-377681267) and followup comments indicate that the second case (float to float) was never UB in the first place - should the second example be removed?

Yes, I think we can remove that altogether -- would you like to do that here?

martin-t:
Sure, I also changed the phrasing into past tense in the first example and removed the bullet point since it's no longer a list. Is is ok like this?

cuviper:
Works for me, thanks!

bors r+

PR #195, opened 2020-11-13
Add support for euclidean division and modulation

### LINKED ISSUE #159: Add support for euclidean division and modulation
Support for `div_euclid` and `rem_euclid` would be very useful, particularly for `vek`: https://github.com/yoanlcq/vek/issues/11.

SparrowLii:
Fixes #159 
Add support for euclidean division and modulation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,7 +40,7 @@ pub use int::PrimInt;
 pub use ops::checked::{
     CheckedAdd, CheckedDiv, CheckedMul, CheckedNeg, CheckedRem, CheckedShl, CheckedShr, CheckedSub,
 };
-pub use ops::euclid::{CheckedDivRemEuclid, DivRemEuclid};
+pub use ops::euclid::{CheckedDivEuclid, CheckedRemEuclid, DivEuclid, RemEuclid};
 pub use ops::inv::Inv;
 pub use ops::mul_add::{MulAdd, MulAddAssign};
 pub use ops::saturating::{Saturating, SaturatingAdd, SaturatingMul, SaturatingSub};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,7 +48,7 @@ pub use ops::wrapping::{
 };
 pub use pow::{checked_pow, pow, Pow};
 pub use sign::{abs, abs_sub, signum, Signed, Unsigned};
-
+pub use ops::euclid::{DivRemEuclid,CheckedDivRemEuclid};
 #[macro_use]
 mod macros;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,6 +40,7 @@ pub use int::PrimInt;
 pub use ops::checked::{
     CheckedAdd, CheckedDiv, CheckedMul, CheckedNeg, CheckedRem, CheckedShl, CheckedShr, CheckedSub,
 };
+pub use ops::euclid::{CheckedDivRemEuclid, DivRemEuclid};
 pub use ops::inv::Inv;
 pub use ops::mul_add::{MulAdd, MulAddAssign};
 pub use ops::saturating::{Saturating, SaturatingAdd, SaturatingMul, SaturatingSub};
@@ -48,7 +49,6 @@ pub use ops::wrapping::{
 };
 pub use pow::{checked_pow, pow, Pow};
 pub use sign::{abs, abs_sub, signum, Signed, Unsigned};
-pub use ops::euclid::{DivRemEuclid,CheckedDivRemEuclid};
 #[macro_use]
 mod macros;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -33,6 +33,7 @@ pub use bounds::Bounded;
 #[cfg(any(feature = "std", feature = "libm"))]
 pub use float::Float;
 pub use float::FloatConst;
+pub use float::FloatCore;
 // pub use real::{FloatCore, Real}; // NOTE: Don't do this, it breaks `use num_traits::*;`.
 pub use cast::{cast, AsPrimitive, FromPrimitive, NumCast, ToPrimitive};
 pub use identities::{one, zero, One, Zero};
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -48,7 +49,6 @@ pub use ops::wrapping::{
 };
 pub use pow::{checked_pow, pow, Pow};
 pub use sign::{abs, abs_sub, signum, Signed, Unsigned};
-
```
cuviper, line 51: nit: this removal is unrelated, and I'd prefer to have that line for separation.

src/ops/euclid.rs
```diff
@@ -0,0 +1,327 @@
+use core::ops::{Div, Rem};
+pub trait DivEuclid: Sized + Div<Self, Output = Self> {
```
cuviper, line 2: Please add a blank line around the trait declarations.

src/ops/euclid.rs
```diff
@@ -0,0 +1,327 @@
...
+    /// such that `self >= n * v`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use num_traits::DivEuclid;
+    ///
+    /// let a: i32 = 7;
+    /// let b: i32 = 4;
+    /// assert_eq!(DivEuclid::div_euclid(a,b), 1); // 7 > 4 * 1
```
cuviper, line 17: nit: this should be spaced `(a, b)` -- and there's a lot of similar non-standard formatting in places that `rustfmt` doesn't reach within docs and macros. Please go through and clean that up if you can.

src/ops/euclid.rs
```diff
@@ -0,0 +1,327 @@
...
+    /// let b: i32 = 4;
+    /// assert_eq!(RemEuclid::rem_euclid(a,b), 3);
+    /// assert_eq!(RemEuclid::rem_euclid(-a,b), 1);
+    /// assert_eq!(RemEuclid::rem_euclid(a,-b), 3);
+    /// assert_eq!(RemEuclid::rem_euclid(-a,-b), 1);
+    /// ```
+    fn rem_euclid(self, v: Self) -> Self;
+}
+macro_rules! div_euclid_int_impl {
+    ($trait_name:ident for $($t:ty)*) => {$(
```
cuviper, line 51: The parameterized `$trait_name` isn't helping us here. I think this should hardcode the trait name and combine `div`/`rem`, so it can be simply invoked `euclid_int_impl(i8 i16 i32 i64 isize)`.

src/ops/euclid.rs
```diff
@@ -0,0 +1,327 @@
...
+        impl $trait_name for $t {
+            #[inline]
+            fn rem_euclid(self, v: $t) -> Self {
+                self % v
+            }
+        }
+    )*}
+}
+div_euclid_int_impl!(DivEuclid for i8 i16 i32 i64);
+div_euclid_uint_impl!(DivEuclid for isize usize u8 u16 u32 u64);
```
cuviper, line 104: `isize` belongs in the signed group.

cuviper (changes requested):
I think we can use `FloatCore` to avoid the `cfg` limitations. Otherwise, there are some style issues, but I think this looks pretty good!

SparrowLii:
Thanks very much for reviewing. I have made the corresponding changes. 
As a Rust learner (even an open source learner)， I look forward to contributing to Rust.

src/ops/euclid.rs
```diff
@@ -0,0 +1,290 @@
...
+    /// assert_eq!(Euclid::rem_euclid(a, -b), 3);
+    /// assert_eq!(Euclid::rem_euclid(-a, -b), 1);
+    /// ```
+    fn rem_euclid(self, v: Self) -> Self;
+}
+macro_rules! euclid_int_impl {
+    ($($t:ty)*) => {$(
+        impl Euclid for $t {
+            #[inline]
+            fn div_euclid(self, v: $t) -> Self {
```
zesterer, line 54: Any reason for this to not use the corresponding method on all integer types?

SparrowLii: That's what I did before.  But this will cause the test of version 1.8.0 to fail, as it shows that the parameter type of div_eculid/rem_eculid is incorrect (expecting &self, not self, or on the contrary). So I implemented a separate implementation to avoid these compatibility issues.

XAMPPRocky:
@cuviper Is there anything left in this PR that needs to be changed? We'd really like this functionality for `rust-gpu` so that we can support `rem_euclid` on SPIR-V platforms. 🙂

cuviper (changes requested):
I think the methods should take references, for the sake of `num-bigint` where it's a waste of allocations to consume values. Either that, or go fully generic with a type parameter RHS and and associated `type Output`, so they can be implemented for any mix of values and references. There are annoyance trade-offs both ways.

SparrowLii:
I think it is reasonable to use reference types from the aspect of efficiency.  If a general type is used and the Output type is specified, we need to implement it separately for the several cases(self, &self, rhs, &rhs). This should probably be what a direct user-facing library should do. As a basic library, num-trait should maintain its concise style IMO

src/lib.rs
```diff
@@ -33,13 +33,15 @@ pub use bounds::Bounded;
 #[cfg(any(feature = "std", feature = "libm"))]
 pub use float::Float;
 pub use float::FloatConst;
+pub use float::FloatCore;
```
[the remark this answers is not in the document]
SparrowLii: Thanks for explanation, has removed it.

src/ops/euclid.rs
```diff
@@ -0,0 +1,296 @@
...
+    #[test]
+    fn euclid_float() {
+        macro_rules! test_euclid {
+            ($($t:ident)+) => {
+                $(
+                    {
+                        let x: $t = 12.1;
+                        let y: $t = 3.2;
+                        assert!(Euclid::div_euclid(&x, &y) * y + Euclid::rem_euclid(&x, &y) - x
+                        <=46.4 * <$t as ::float::FloatCore>::epsilon());
```
cuviper, line 264:
```suggestion
                        <= 46.4 * <$t as ::float::FloatCore>::epsilon());
```

cuviper (commented):
Thanks, it looks nice.

Maybe we should also add a combined `div_rem_euclid`? (and checked.) That would let you get both values with the divisions and conditions only checked once. All that probably gets optimized together with primitive values anyway, but it would matter for types like `BigInt`.

PR #199, opened 2021-01-08
Export trait implementation macros

cargodog:
The trait implementation macros are quite useful for other crates which may choose to implement numeric types. This patch exports those macros to facilitate this.

[Here](https://github.com/dalek-cryptography/curve25519-dalek/blob/e69be46e22ca5e8ba5aae2cdbdc886d1d2db55e8/src/scalar.rs#L1143) is an example scenario where it would be quite useful to have access to these macros down stream.

### DIFF
--- a/src/bounds.rs
+++ b/src/bounds.rs
@@ -14,6 +14,7 @@ pub trait Bounded {
     fn max_value() -> Self;
 }
 
+#[macro_export]
 macro_rules! bounded_impl {
     ($t:ty, $min:expr, $max:expr) => {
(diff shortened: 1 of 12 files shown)


cuviper:
These macros have been developed just "good enough" for internal use, not with an eye toward permanent stability. I'd rather take it case-by-case and see if something useful should be added to `num-derive` instead.

cargodog:
Ah yes, that would be a better approach. Thanks for your time. I'll go ahead and close this out

PR #201, opened 2021-02-05
Use standard library for float from_str_radix with radix 10

### LINKED ISSUE #198: Rounding error with `f64::from_str_radix`
Consider the following program:
```rust
use num::Num;

fn main() {
    let y = "2.89".parse::<f64>();
    println!("{:?}", y);
    let x = f64::from_str_radix("2.89", 10);
    println!("{:?}", x);
}
```
It prints:
```
Ok(2.89)
Ok(2.8899999999999997)
```
Am I missing anything here or results should be equal?

tspiteri:
Fixes #198.

While at it, also make parsing of the `"-NaN"` string match the standard library.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -236,6 +236,7 @@ macro_rules! float_trait_impl {
                     "inf"   => return Ok(core::$t::INFINITY),
                     "-inf"  => return Ok(core::$t::NEG_INFINITY),
                     "NaN"   => return Ok(core::$t::NAN),
+                    "-NaN"  => return Ok(-core::$t::NAN),
                     _       => {},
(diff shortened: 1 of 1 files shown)


tspiteri:
https://github.com/rust-num/num-traits/issues/198#issuecomment-708512381 says "it would be nicer to improve this for all bases", but that is a lot of work with basically zero usefulness, which is [the reason `from_str_radix` was removed from the standard library floating-point numbers](https://internals.rust-lang.org/t/deprecate-f-32-64-from-str-radix/2405).

cuviper:
Thanks!

@bors r+

PR #202, opened 2021-02-12
PrimInt: add reverse_bits() method

Xiretza:


cuviper:
Unfortunately, it's a breaking change to add a new trait method without a default implementation, because we haven't prevented downstream crates from implementing this trait, and they would fail without this new addition.

Your implementation also requires Rust 1.37.

Xiretza:
I see, that's unforunate. Are there any plans for a new major version (or, since it's still at 0.x, any new version that includes breaking changes)?

tspiteri:
It is possible to write a default implementation of `reverse_bits` which only depends on `PrimInt` methods. For example this works for me and it also gives me similar assembly on x86_64 to the new primitive `reverse_bits` methods (the only difference I could spot is in instruction order).
```rust
fn one_per_byte<P: PrimInt>() -> P {
    let bytes = P::zero().count_zeros() / 8;
    // if bytes == 1, return 0x01
    // if bytes == 2, return 0x0101 = (0x01 << 8) | 0x01
    // if bytes == 4, return 0x01010101 = (0x0101 << 16) | 0x0101
    let mut ret = P::one();
    let mut shift = 8;
    let mut b = bytes >> 1;
    while b != 0 {
        ret = (ret << shift) | ret;
        shift <<= 1;
        b >>= 1;
    }
    ret
}
fn reverse_bits<P: PrimInt>(i: P) -> P {
    let rep_01: P = one_per_byte();
    let rep_03 = (rep_01 << 1) | rep_01;
    let rep_05 = (rep_01 << 2) | rep_01;
    let rep_0f = (rep_03 << 2) | rep_03;
    let rep_33 = (rep_03 << 4) | rep_03;
    let rep_55 = (rep_05 << 4) | rep_05;

    // code above only used to determine rep_0f, rep_33, rep_55;
    // optimizer should be able to do it in compile time
    let mut ret = i.swap_bytes();
    ret = ((ret & rep_0f) << 4) | ((ret >> 4) & rep_0f);
    ret = ((ret & rep_33) << 2) | ((ret >> 2) & rep_33);
    ret = ((ret & rep_55) << 1) | ((ret >> 1) & rep_55);
    ret
}
```

Xiretza:
I added @tspiteri's fallback behind an autocfg gate, but I haven't been able to confirm whether it actually works as designed - I can't find any traces of the fallback method in the output binary (with stable rustc) even if I delete the method in the trait implementation entirely.

build.rs
```diff
@@ -16,5 +16,7 @@ fn main() {
         "has_to_int_unchecked",
     );
 
+    ac.emit_path_cfg("u32::reverse_bits()", "has_reverse_bits");
```
cuviper, line 19: This test is failing (see `cargo check -vv`) because it will try `pub use PATH;` which doesn't work for inherent methods, nevermind the parentheses. You could try an expression instead, like:
```suggestion
    ac.emit_expression_cfg("1u32.reverse_bits()", "has_reverse_bits");
```

Xiretza: Ah, thanks! Good to know that making cargo more noisy outputs some autocfg information.

I don't know how rust libraries work in detail, but the only change I can see in `libnum_traits.rlib` with and without the fallback is in `lib.rmeta`, not sure if that's ok.

cuviper (commented):
This is good, but I'd like to see more tests, especially to cover signed/unsigned and different sizes for the fallback implementation. (Probably just unit tests, not cluttering the doc tests.)

Xiretza:
Sorry for the noise, I kept running into CI issues that I didn't have locally.

cuviper:
Thanks!

bors r+

PR #203, opened 2021-02-12
A few small patches

sshilovsky:


cuviper:
Thanks, and sorry for the horrible review delay.

bors r+

PR #205, opened 2021-03-14
Add leading_ones and trailing_ones to PrimInt

clarfonthey:
Since this was only stabilised in 1.46.0, it falls back to a naïve version on older versions.

It seems unlikely that the version in libstd will ever be different from calling the zeros versions on !self, but for future-proofing, this defers to the libstd versions anyway.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -16,5 +16,7 @@ fn main() {
         "has_to_int_unchecked",
     );
 
+    ac.emit_expression_cfg("1u32.trailing_ones()", "has_leading_trailing_ones");
+
     autocfg::rerun_path("build.rs");
 }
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -16,5 +16,7 @@ fn main() {
         "has_to_int_unchecked",
     );
 
+    ac.emit_path_cfg("1.trailing_ones()", "has_leading_trailing_ones");
```
cuviper, line 19: This fails even on current stable -- see `cargo check -vv`. This needs to be an expression check using an explicit integer type to pass.

clarfonthey: Ah, whoops. Good catch.

clarfonthey:
Rebased and fixed issue mentioned

clarfonthey:
ping @cuviper, I think you'd be the one to merge this?

Issue #189, opened 2020-10-01
PrimInt: to_be_bytes & co


chrysn:
The current `to_be` functions require unsafe to get a slice out of them; the individual types now (since 1.32) have `to_be_bytes` that does not have that flaw.

I suggest adding `to_be_bytes()` as well as `to_ne_bytes` and `to_le_bytes` as well as the analogous `from_be_bytes`, `from_ne_bytes` and `from_le_bytes`.

To implement this, `PrimInt` would need to get an associated type that reflects such a buffer; depending on when this gets implemented and what MSRV is targeted then, that could be a `const NBYTES: usize` (where to_be_bytes would return `[u8; NBYTES]`), or a `BytesBuffer` associated type with the right array lengths. In the latter form, the associated type would be required by the trait to implement at least `TryFrom<&[u8]>` and `AsRef<&[u8]>`.

Researching into whether that latter form would be upgradable to a const generic one easily later, I found that `PrimInt` is not sealed; that didn't sound too bad until I noticed that even a default implementation that uses `to_be()` and then transmutes around could still not come up with a suitable sized and owned return value short of returning a `Box<[u8]>` (that would technically satisfy the bounds but be ugly). Worse yet, it'd slam the door on being able to migrate towards returning `[u8; NBYTES]` later (which otherwise could be API compatible as the new return type does satisfy all bounds previously placed on it).

Is there any evolution mechanism in num-traits through which `to_be_bytes()` could be added?

cuviper:
I don't know any way to extend `PrimInt` for this.

There could be a new trait though -- #103 was working on this, but it would need to be rebased and reviewed again. Someone can also come up with a new design, if they prefer.

kaidokert:
Is #103 still the current thinking around it ? If yes, i might want to pick it up and try to resolve pending issues. Related question, if [OverflowingShl/Shr implementation](https://github.com/kaidokert/fixed-bigint-rs/blob/main/src/patch_num_traits.rs) like this would also be accepted ?

ctrlcctrlv:
Is this crate simply abandoned? I need the floating point versions of these (`to_bits`, `from_bits` specifically, stabilized in 1.20; bytes stabilized in 1.40: https://github.com/rust-lang/rust/commit/7b72c2831c6a50a9e07a3d94ebef5d9f29bfd51e).

[...]

PR #207, opened 2021-03-17
Add copysign

### LINKED ISSUE #152: Float trait doesn't require copysign function

XAMPPRocky:
Resolves #152
Closes #155

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,6 @@ libm = { version = "0.2.0", optional = true }
 default = ["std"]
 std = []
 i128 = []
-copysign = []
 
 [build-dependencies]
 autocfg = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ libm = { version = "0.2.0", optional = true }
 default = ["std"]
 std = []
 i128 = []
+copysign = []
 
 [build-dependencies]
 autocfg = "1"
--- a/build.rs
+++ b/build.rs
@@ -11,6 +11,10 @@ fn main() {
         autocfg::emit("has_i128");
     }
 
+    if env::var_os("CARGO_FEATURE_COPYSIGN").is_some() || ac.probe_expression("f32::copysign") {
+        autocfg::emit("has_copysign");
+    }
+
     ac.emit_expression_cfg(
         "unsafe { 1f64.to_int_unchecked::<i32>() }",
         "has_to_int_unchecked",
--- a/build.rs
+++ b/build.rs
@@ -11,10 +11,6 @@ fn main() {
         autocfg::emit("has_i128");
     }
 
-    if env::var_os("CARGO_FEATURE_COPYSIGN").is_some() || ac.probe_expression("f32::copysign") {
-        autocfg::emit("has_copysign");
-    }
-
     ac.emit_expression_cfg(
         "unsafe { 1f64.to_int_unchecked::<i32>() }",
         "has_to_int_unchecked",
@@ -24,6 +20,10 @@ fn main() {
     ac.emit_expression_cfg("1u32.trailing_ones()", "has_leading_trailing_ones");
     ac.emit_expression_cfg("{ let mut x = 1; x += &2; }", "has_int_assignop_ref");
     ac.emit_expression_cfg("1u32.div_euclid(1u32)", "has_div_euclid");
+
+    if env::var_os("CARGO_FEATURE_STD").is_some() {
+        ac.emit_expression_cfg("1f64.copysign(-1f64)", "has_copysign");
+    }
 
     autocfg::rerun_path("build.rs");
 }
--- a/src/float.rs
+++ b/src/float.rs
@@ -2314,21 +2314,21 @@ mod tests {
 
     #[cfg(any(feature = "std", feature = "libm"))]
     fn test_copysignf(p: f32, n: f32, nan: f32) {
-        use float::Float;
         use core::ops::Neg;
+        use float::Float;
 
         assert!(p.is_sign_positive());
         assert!(n.is_sign_negative());
         assert!(nan.is_nan());
 
-        assert_eq!(p, p.copysign(p));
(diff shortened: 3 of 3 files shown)


cuviper:
The standard library didn't get `copysign` until 1.35. In CI for earlier versions, you're hitting `#![deny(unconditional_recursion)]`, as the attempted macro-forwarding actually ends up calling the trait method again, instead of the newer inherent method. You'll need to add a `build.rs` check for the method, and then `cfg` the forwarding based on that, leaving it to the default impl otherwise.

XAMPPRocky:
> You'll need to add a build.rs check for the method, and then cfg the forwarding based on that, leaving it to the default impl otherwise.

I think I've implemented that, but it's still failing the tests, and I still can't really tell what I need to do to fix it.

src/float.rs
```diff
@@ -1917,6 +1946,11 @@ macro_rules! float_impl_std {
                 Self::acosh(self) -> Self;
                 Self::atanh(self) -> Self;
             }
+
+            #[cfg(has_copysign)]
+            forward! {
+                Self::copysign(self, sign: Self) -> Self;
+            }
```
tspiteri, line 1953: Hmm, 1.8.0 seems to expand this even if not `cfg(has_copysign)`. While it correctly leaves it out if the function is written directly instead of using the `forward!` macro. That is it correctly leaves out
```rust
            #[cfg(has_copysign)]
            #[inline]
            fn copysign(self, sign: Self) -> Self {
                Self::copysign(self, sign)
            }
```

After some investigation, it seems to me that this bug does not affect a `#[cfg(...)] forward!{...}` that is not inside another macro, so this is the only place I could find in the file. So simply expanding this single instance of the `forward!` macro should fix the build I think.

XAMPPRocky: Thanks for the investigation, 1.8.0 seems to just not work on my MacBook so I wasn't able to test 1.8.0. I've updated the code with your suggestion.

XAMPPRocky:
The workflow needs to be approved :)

XAMPPRocky:
@tspiteri Thanks for that, it seems like it's passing now.

XAMPPRocky:
@cuviper Friendly review ping 🙂

Cargo.toml
```diff
@@ -23,6 +23,7 @@ libm = { version = "0.2.0", optional = true }
 default = ["std"]
 std = []
 i128 = []
+copysign = []
```
cuviper, line 26: Do we really need a feature for this? I'd rather leave it to `autocfg`, unless there's a strong reason to expose that condition.

build.rs
```diff
@@ -11,6 +11,10 @@ fn main() {
         autocfg::emit("has_i128");
     }
 
+    if env::var_os("CARGO_FEATURE_COPYSIGN").is_some() || ac.probe_path("f32::copysign") {
```
cuviper, line 14: `f32::copysign` is not a valid path, or at least it's not one that works with `use PATH` as attempted by `probe_path`. You should be able to `probe_expression` though.

(`cargo check -vv` is helpful to see build script output -- this path is failing even on current stable.)

src/float.rs
```diff
@@ -2243,4 +2283,28 @@ mod tests {
         check::<f32>(1e-6);
         check::<f64>(1e-12);
     }
+
+    #[cfg(all(any(feature = "std", feature = "libm"), has_copysign))]
```
cuviper, line 2287: Why does the test need `std`/`libm`? Shouldn't the default implementation work too?

Cargo.toml
XAMPPRocky: Yes, because `autocfg`  is not something that can be used on custom targets (related https://github.com/rust-lang/cargo/issues/7501). I would go as far as saying that `autocfg` is an anti-pattern for supporting custom targets currently, the way probing capabilities currently works is very incompatible if your project requires being built in a specialised way. One example a project that needs `num-traits` and a custom target is `rust-gpu`.

cuviper: Yeah, even as the `autocfg` author I am frustrated by that situation, but I haven't found time to design/push for tool changes that would alleviate it. I think having something like `TARGET_RUSTFLAGS` from cargo set for build scripts would solve a lot of issues. Might you be interested in pursuing this, since it impacts targets you work with?

src/float.rs
cuviper: Oh, of course, that was gated because `Float` itself has to exist. Sorry for the misdirection. And the default impl will still be tested on older `std`-only CI.

src/float.rs
```diff
@@ -2095,6 +2131,8 @@ impl Float for f64 {
         libm::atanh as atanh(self) -> Self;
         libm::fmax as max(self, other: Self) -> Self;
         libm::fmin as min(self, other: Self) -> Self;
+        libm::copysign as copysign(self, sign: Self) -> Self;
+        libm::copysignf as copysignf(self, sign: Self) -> Self;
```
cuviper, line 2135: We only need the first one here. :)

src/float.rs
cuviper: Shouldn't need to check `has_copysign` though.

ctrlcctrlv:
(Merged into https://github.com/MFEK/num-traits.rlib/commit/ce9a7dfc17ba80bb1489eda680694f01cbd13f84 due to https://github.com/rust-num/num-traits/issues/189#issuecomment-1011828593.)

Cargo.toml
cuviper: FWIW, `autocfg` uses `CARGO_ENCODED_RUSTFLAGS` now, which should be more reliable.

If you're not even happy with that, then I'd prefer a plain version check rather than a feature for every little implementation detail. ("i128" is a little different because it actually adds to the API.)

Issue #208, opened 2021-03-18
Some way to combine Bounded and Unsigned + Zero


clarfonthey:
Right now it would be a breaking change to split up the `Bounded` trait, but I at least have one use case for this after messing around with AltSysrq/proptest#235: there are types that are lower-bounded but not upper-bounded, specifically `BigUint`. Would be nice to be able to have a trait that lets you get a min_value without requiring a max_value, since right now `Bounded` and `Unsigned + Zero` may overlap and we can't rely on that.

cuviper:
I'm surprised there's a scenario where a possibly-negative minimum would be useful without also caring about a maximum. Can you show an example of how you would use this?

Maybe we could have a trait that returns `Option<Self>` bounds? But I'm not sure if there's any type that would have an upper bound and not lower, so maybe a simple `LowerBounded` trait is enough.

clarfonthey:
In this particular case, it was because `RangeTo` can be converted to a `Range` for unsigned integers and bounded integers, but `RangeFull` cannot.

clarfonthey:
Closing this issue since the PR I merged accounts for this.

PR #210, opened 2021-03-22
Add LowerBounded/UpperBounded traits

clarfonthey:
Potential solution for #208. With a breaking change, these could simply become supertraits of `Bounded`, but until then, we have to deal with blanket implementations.

I added both simply because it was easy to do so, although we could opt to not require both. I don't see it being that negative to include both, however.

### DIFF
--- a/src/bounds.rs
+++ b/src/bounds.rs
@@ -8,10 +8,36 @@ use core::{u16, u32, u64, u8, usize};
 /// Numbers which have upper and lower bounds
 pub trait Bounded {
     // FIXME (#5527): These should be associated constants
-    /// returns the smallest finite number this type can represent
+    /// Returns the smallest finite number this type can represent
     fn min_value() -> Self;
-    /// returns the largest finite number this type can represent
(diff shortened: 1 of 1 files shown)


regexident:
Both `trait LowerBounded` and `trait UpperBounded` on their own look like a reasonable addition to `num-traits`. In light of the existing `trait Bounded` however they —and especially their blanket implementations— feel wrong to me as `trait Bounded` —like you said— should clearly be redefined in terms of `trait Bounded: LowerBounded + UpperBounded {}`, not the other way round.

The problem these traits aim to address is one of [many arguments](https://github.com/rust-num/num-traits/issues/47) in favor of going thru with a future breaking release, while the proposed compromise/workaround might end up working against just that.

@cuviper this deserves an entry in the [tracking issue](https://github.com/rust-num/num-traits/issues/47), I think.

cuviper:
I added that to the tracking issue. Otherwise, this looks fine to me, thanks!

bors r+

PR #211, opened 2021-03-24
update readme for rust 2018 edition

ibraheemdev:


cuviper:
We still support older versions, but I guess anyone doing so should know about `extern crate` by now. So for current new users, it seems fine to drop that advice. Thanks!

bors r+

PR #214, opened 2021-04-15
Ignore case for float parsing of text special values

tspiteri:
This copies some of the standard library fixes of <https://github.com/rust-lang/rust/pull/78618>.


cuviper:
Thanks!

bors r+

PR #218, opened 2021-06-24
Implement is_sign_* and signum methods in terms of bitcasts.

ElectronicRU:
# Rationale

[rust-gpu](/EmbarkStudios/rust-gpu) project uses num-traits directly and via glam. However, integer_decode() function is a bit poisonous for it - it immediately requires both Int16 and Int64 capabilities.

This PR reimplements corresponding functions in terms of libm's copysign. (For some reason, rust's libm is missing signbit - it can be done with a bitcast though, maybe that's actually a better implementation avenue).

### DIFF
--- a/src/float.rs
+++ b/src/float.rs
@@ -783,6 +783,17 @@ impl FloatCore for f32 {
 
     #[inline]
     #[cfg(not(feature = "std"))]
+    fn is_sign_negative(self) -> bool {
+        const SIGN_MASK: u32 = 0x80000000;
+
+        // Safety: this identical to the implementation of f32::to_bits(),
+        // which is only available starting at Rust 1.20
+        let bits: u32 = unsafe { mem::transmute(self) };
+        bits & SIGN_MASK != 0
+    }
+
+    #[inline]
+    #[cfg(not(feature = "std"))]
     fn to_degrees(self) -> Self {
         // Use a constant for better precision.
         const PIS_IN_180: f32 = 57.2957795130823208767981548141051703_f32;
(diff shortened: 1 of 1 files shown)


cuviper:
> However, integer_decode() function is a bit poisonous for it - it immediately requires both Int16 and Int64 traits.

Can you elaborate? I don't understand what's poisonous, or what those traits are.

ElectronicRU:
> > However, integer_decode() function is a bit poisonous for it - it immediately requires both Int16 and Int64 traits.
> 
> Can you elaborate? I don't understand what's poisonous, or what those traits are.

Simply put, those traits (capabilities actually - I mistyped) are what is required of target hardware to successfully translate spir-v to machine code. Nonstandard integer sizes present a problem, especially on older hardware.

ElectronicRU:
If there are worries of performance, I can re-do it as a separate feature flag, depending on `libm`. Would that be necessary or more agreeable?

cuviper:
I wasn't really thinking about performance yet, but I guess it is a little unfortunate that the `libm` call won't usually be inlined, unless you enable LTO.

Maybe we should just write the no-`std` version of `is_sign_negative` directly, with only that small subset of `integer_decode_f32`/`f64` that deals with the sign bit. Then `signum` should still be fine with the default impl.

ElectronicRU:
Fair enough!

cuviper:
Thanks!

@bors r+

cuviper:
bah, wrong bot...

bors r+

PR #229, opened 2021-12-15
feat: self-referential bounded trait for generics

samuelhnrq:
I was using this crate to recieve a generic number T bounded by this create's traits more specifically the 'Bounded' trait since I needed the upper bounds for some audio calculations, (wav sample format conversion) but I only have the instance of a number and would be much more convenient to have a "self referenced" edititon of the  `Bounded::max_value`  method to enable stuff like:

```rust
fn unwrap_sample<Sample>(sample_result: Result<Sample, hound::Error>) -> f32
where
  Sample: Num + Bounded + ToPrimitive,
{
  let sample_val = sample_result.unwrap();
  let max = sample_val.own_max_value();
  return sample_val.div(max).to_f32().unwrap();
}
```

### DIFF
--- a/src/bounds.rs
+++ b/src/bounds.rs
@@ -10,8 +10,10 @@ pub trait Bounded {
     // FIXME (#5527): These should be associated constants
     /// Returns the smallest finite number this type can represent
     fn min_value() -> Self;
+    fn own_min_value(&self) -> Self;
     /// Returns the largest finite number this type can represent
     fn max_value() -> Self;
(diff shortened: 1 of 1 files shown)


samuelhnrq:
Didn't give much thought to the name, no hard feelings if you find a more suitable name

cuviper:
> ```rust
> let max = sample_val.own_max_value();
> ```

Why not just `Sample::max_value()` here?

PR #244, opened 2022-08-06
Atomic Traits

zommiommy:
Hi! I've added my implementation of `Atomic` and `IntoAtomic`, I use them really often and they are quite similar to your traits so I think that they might be a good fit in your library.
I still have to write documentation and examples, but I wanted to know what you think about this.

Also, we could add the opt-in support for the methods from the unstable feature `#![feature(atomic_from_mut)]` by adding an enabling feature to the create.

### DIFF
--- a/src/atomics.rs
+++ b/src/atomics.rs
@@ -32,6 +32,7 @@ pub trait Atomic: Sized {
     ) -> Self::NonAtomicType;
 
     fn fetch_add(&self, value: Self::NonAtomicType, order: Ordering) -> Self::NonAtomicType;
+    fn fetch_saturating_add(&self, value: Self::NonAtomicType, order: Ordering) -> Self::NonAtomicType;
     fn fetch_and(&self, value: Self::NonAtomicType, order: Ordering) -> Self::NonAtomicType;
     fn fetch_max(&self, value: Self::NonAtomicType, order: Ordering) -> Self::NonAtomicType;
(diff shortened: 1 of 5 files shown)


src/atomics.rs
```diff
@@ -0,0 +1,215 @@
+use core::sync::atomic::*;
+
+pub trait Atomic: Sized {
+    type NonAtomicType: Copy;
```
qinsoon, line 4: I came across this pull request while I am trying to solve a similar problem. It would be better the `NonAtomicType` is a `PrimInt` (or whatever num trait that is more suitable). In that case, for `NonAtomicType` in the arguments and return values, you can use num traits methods on them.

zommiommy: Yeah, that's probably a sensible change

src/coerced.rs
```diff
@@ -0,0 +1,30 @@
+pub trait Coerced<F> {
```
cuviper, line 1: I think this means the same as [`trait AsPrimitive`](https://docs.rs/num-traits/latest/num_traits/cast/trait.AsPrimitive.html)?

cuviper:
Hi! I don't have a lot of time for working on this crate, and I think atomics are different enough that this would do well as a crate of its own. You could even call it `num-atomic` if you like -- I don't own the crate prefix. :)

PR #246, opened 2022-08-14
add BITS associated constant to PrimInt

umanwizard:


cuviper:
Unfortunately, this is a breaking change for external implementors, especially since this trait isn't sealed.

It's tempting to add a default to `8 * mem::size_of::<Self>()`, but nothing says that the entire value has to be numeric bits either. It could be a wrapper with some other extraneous metadata, for example.

#247 suggests a new trait, which would avoid a breaking change.

PR #251, opened 2022-11-02
Add `CheckedSum` and `CheckedProduct` traits

### LINKED ISSUE #250: Add `CheckedSum` and `CheckedProduct` traits
The Rust language team decided not to add `CheckedSum` and `CheckedProduct` traits to core https://github.com/rust-lang/rust/pull/95485 but it was suggested that they be added here.  

These traits allow you to add (or multiply) together all numbers in an iterator but return `None` instead of panicking or wrapping if an overflow occurs.

I think the best way to do it is with a blanket implementation:

```rust
pub trait CheckedSum<A = Self> : Sized {
    fn checked_sum<I: Iterator<Item = A>>(iter: I) -> Option<Self>;
}

impl<T> CheckedSum<T> for T where T : CheckedAdd + Sized + crate::identities::Zero
{
    fn checked_sum<I: Iterator<Item = Self>>(iter: I) -> Option<Self> {
        let mut total = Self::zero();
        for i in iter{
            total = total.checked_add(&i)?;
        }
        Some(total)
    }
}
```

I will submit a pull request and if that is accepted I might do the same for Overflowing, Saturating, and Wrapping

wainwrightmark:
Closes https://github.com/rust-num/num-traits/issues/250

This adds `CheckedSum` and `CheckedProduct` traits as well as blanket implementations for them. 

 It also adds `CheckedSumIter` and `CheckedProductIter` traits and blanket implementations for those as that provides a much more ergonomic interface, allowing you to do:

```rust
asserteq!(Some(42), [40,2].iter().checked_sum());
asserteq!(Some(42), [40,2].into_iter().checked_sum());
```

### DIFF
--- a/src/iter/checked_product.rs
+++ b/src/iter/checked_product.rs
@@ -1,15 +1,15 @@
 use crate::identities::One;
 use crate::CheckedMul;
 
-/// This trait represents types of which an iterator can be multiplied up with overflow checking.
+/// Trait to represent types that can be created by multiplying elements of an iterator with overflow checking.
 /// This trait should rarely be called directly.
-pub trait CheckedProduct<Result = Self> {
-    /// Multiplies up the elements of an iterator, returning `None` if an overflow would occur.
+pub trait CheckedProduct<A = Self> : Sized {
+    /// Method which takes an iterator and generates Self from the elements by multiplying the items with overflow checking, returning `None` if the multiplication would overflow.
     ///
-    /// Multiplies up an empty iterator returns a value representing One.
+    /// An empty iterator returns the one value of the type.
     ///
-    /// If the iterator contains Zero, the order of elements may effect whether the result is `None`.
-    fn checked_product<I: Iterator<Item = Self>>(iter: I) -> Option<Result>;
+    /// For iterators containing zero, the order of elements may effect whether the result is `None`.
+    fn checked_product<I: Iterator<Item = A>>(iter: I) -> Option<Self>;
 }
 
 impl<T> CheckedProduct<T> for T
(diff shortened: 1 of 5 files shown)


src/iter/checked_product.rs
```diff
@@ -0,0 +1,95 @@
...
+
+/// This trait represents types of which an iterator can be multiplied up with overflow checking.
+/// This trait should rarely be called directly.
+pub trait CheckedProduct<Result = Self> {
+    /// Multiplies up the elements of an iterator, returning `None` if an overflow would occur.
+    ///
+    /// Multiplies up an empty iterator returns a value representing One.
+    ///
+    /// If the iterator contains Zero, the order of elements may effect whether the result is `None`.
+    fn checked_product<I: Iterator<Item = Self>>(iter: I) -> Option<Result>;
```
cuviper, line 12: This is opposite of how `Product` is defined, where the type parameter is the _input_ (`Item`) and it always outputs `Self`. I would rather match that for consistency, and same for `Sum`.

src/iter/checked_product.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+///This trait is for iterators that can be multiplied up with overflow checking.
+trait CheckedProductIter<T, Result>: Iterator<Item = T> {
+    /// Multiplies up the elements of an iterator, returning `None` if an overflow would occur.
+    ///
+    /// Multiplies up an empty iterator returns a value representing One.
+    ///
+    /// If the iterator contains Zero, the order of elements may effect whether the result is `None`.
+    fn checked_product(self) -> Option<Result>;
```
cuviper, line 40: Similarly, for consistency with `Iterator`, this trait should have no parameters, and the method should be parameterized `checked_product<P: CheckedProduct<Self::Item>>`.

I think then we could also combine your two iterator extensions into one `CheckedIterator` trait, or even `NumIterator` to leave room for more extension methods.

cuviper (changes requested):
Rather than a new `iter` module here, I think it would make more sense to move this to the `num-iter` crate.

src/iter/checked_product.rs
wainwrightmark: I completely agree. I feel very silly for having done it the other way around. I have changed it now.

src/iter/checked_product.rs
wainwrightmark: > Rather than a new `iter` module here, I think it would make more sense to move this to the `num-iter` crate.

I wasn't sure. I'm very happy to close this pull request and move the code changes over there if that's best.

PR #258, opened 2023-02-11
Improved documentation.

Enyium:
- `FromPrimitive` functions now refer to implementor for details.

---

See https://github.com/rust-num/num-rational/issues/113.

### DIFF
--- a/src/cast.rs
+++ b/src/cast.rs
@@ -16,39 +16,39 @@ use core::{u128, u16, u32, u64, u8, usize};
 /// even a large `f64` saturating to `f32` infinity.
 pub trait ToPrimitive {
     /// Converts the value of `self` to an `isize`. If the value cannot be
-    /// represented by an `isize`, then `None` is returned.
+    /// represented by an `isize`, `None` is returned.
     #[inline]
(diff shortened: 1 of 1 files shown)


src/cast.rs
```diff
@@ -16,39 +16,39 @@ use core::{u128, u16, u32, u64, u8, usize};
 /// even a large `f64` saturating to `f32` infinity.
 pub trait ToPrimitive {
     /// Converts the value of `self` to an `isize`. If the value cannot be
-    /// represented by an `isize`, then `None` is returned.
```
frewsxcv, line 19: How does removing the word "then" change the meaning of the sentence?

Enyium: It doesn't change the meaning. Seeing all those "then" removals listed, you may think it's something central. But I just changed it at one place for the more common phrasing, and then saw that the sentence was copied elsewhere, so I had to change these occurrences also.

The central piece of the PR is the addition of the lines that say `` See `Self` type or trait implementation for details. `` But since this is just a casual doc comment improvement, I don't think the additional "then" removals are a problem.

Issue #247, opened 2022-08-17
[feature request] Bits trait


kyp44:
Most Rust primitives have a `BITS` associated const (with the number of bits used), but evidently there is no way to use this generically. I have a particular need for this and it's easy enough to implement myself, but it seems like a good candidate for a `num-traits` trait.

kyp44:
Though this is obviously pretty trivial to figure out, here is an example for a subset of primitives.

```
pub trait Bits {
    fn bits() -> u32;
}
impl Bits for u8 {
    fn bits() -> u32 {
        Self::BITS
    }
}
impl Bits for u16 {
    fn bits() -> u32 {
        Self::BITS
    }
}
impl Bits for u32 {
    fn bits() -> u32 {
        Self::BITS
    }
}
impl Bits for u64 {
    fn bits() -> u32 {
        Self::BITS
    }
}
```

PR #271, opened 2023-05-24
Add traits for `BITS`

JonathanWoollett-Light:
Adds a trait for getting the `BITS` constants e.g. [`u32::BITS`](https://doc.rust-lang.org/std/primitive.u32.html#associatedconstant.BITS).

### DIFF
--- a/src/bits.rs
+++ b/src/bits.rs
@@ -1,0 +1,23 @@
+use core::{i128, i16, i32, i64, i8, isize};
+use core::{u128, u16, u32, u64, u8, usize};
+
+/// Numbers which have a specified number of bits.
+pub trait Bits {
+    /// Returns the size of the number in bits.
+    fn bits() -> u32;
+}
+
+macro_rules! bits_impl {
+    ( $($x:ident),* ) => (
+        $(
+            impl Bits for $x {
+                fn bits() -> u32 {
+                    $x::BITS
+                }
+            }
+        )*
+
+    );
+}
+
+bits_impl!(i128, i16, i32, i64, i8, isize, u128, u16, u32, u64, u8, usize);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -32,6 +32,7 @@ pub use crate::bounds::Bounded;
 pub use crate::float::Float;
 pub use crate::float::FloatConst;
 // pub use real::{FloatCore, Real}; // NOTE: Don't do this, it breaks `use num_traits::*;`.
+pub use crate::bits::Bits;
 pub use crate::cast::{cast, AsPrimitive, FromPrimitive, NumCast, ToPrimitive};
 pub use crate::identities::{one, zero, One, Zero};
 pub use crate::int::PrimInt;
@@ -51,6 +52,7 @@ pub use crate::sign::{abs, abs_sub, signum, Signed, Unsigned};
 #[macro_use]
 mod macros;
 
+pub mod bits;
 pub mod bounds;
 pub mod cast;
 pub mod float;


cuviper:
What is your use case? Preferably not a hypothetical, or "because we can," but what is a real example where you would like to use generic `T: Bits`?

See also #246 and #247, but those did not provide justification either...

This is also different than `BigUint::bits(&self)`, for example, which returns the number of *significant* bits as `u64`. If we don't expect such types to implement this trait, we should make that clear.

Also, `{integer}::BITS` have only been stable since Rust 1.53, and this crate maintains an older MSRV, so it would need some conditional handling here. A fallback of `8 * size_of` is probably fine.

JonathanWoollett-Light:
We didn't use this crate in this PR https://github.com/firecracker-microvm/firecracker/pull/3722 because we needed to get the `BITS` field and this wasn't supported. We instead defined our own trait for it https://github.com/firecracker-microvm/firecracker/pull/3722/commits/0b215a58311c3397b8a0eeb5e834bff71e647caa#diff-67fb01ab595aeec3769d00a0a18cc6c71eade58ff873f137fcbaa3802e26d137R114 which is awkward.

---

> What is your use case? Preferably not a hypothetical, or "because we can," but what is a real example where you would like to use generic T: Bits?
> See also https://github.com/rust-num/num-traits/pull/246 and https://github.com/rust-num/num-traits/issues/247, but those did not provide justification either...

I might also suggest this is an incorrect approach, it is inevitable use cases will arise that are not foreseen, and most users will then simply see its not supported and write their own implementation, most people will not post an issue or PR when a library doesn't support what they need, but this is very opinionated and beside the point.

tarcieri:
When using big integers as backing storage for cryptographic keys, there are many times it can be helpful to know the bit size of the key

cuviper:
Is there a reason you made this a function? This PR proposes `fn bits() -> u32`, but:

> We instead defined our own trait for it

... with `const BITS: u32`, which also looks like what RustCrypto is using.

JonathanWoollett-Light:
> Is there a reason you made this a function? This PR proposes `fn bits() -> u32`, but:
> 
> > We instead defined our own trait for it
> 
> ... with `const BITS: u32`, which also looks like what RustCrypto is using.

[`Bounded`](https://docs.rs/num-traits/latest/num_traits/bounds/trait.Bounded.html) uses functions ([`min_value`](https://docs.rs/num-traits/latest/num_traits/bounds/trait.Bounded.html#tymethod.min_value)() and [`max_value`](https://docs.rs/num-traits/latest/num_traits/bounds/trait.Bounded.html#tymethod.max_value)()) to define the max and min values, I used a function to be consistent with this.

I do not have a strong opinion about using a constant or function.

tarcieri:
A function might be a bit more helpful, since the size can be determined at runtime: https://github.com/search?q=repo%3ARustCrypto%2FRSA+bits%28%29&type=code

FWIW we’re looking at adding a runtime-selected fixed-width integer type to `crypto-bigint` for those purposes: https://github.com/RustCrypto/crypto-bigint/pull/221

cuviper:
> `Bounded` uses functions (`min_value` and `max_value`) to define the max and min values,

That trait predates both `const fn` and associated `const` features. Plus those need to return `Self`, which might not have a `const` constructor at all for some types.

> A function might be a bit more helpful, since the size can be determined at runtime

Then that would need to be `fn bits(&self) -> u32` (or larger?), and this is again similar to `BigUint::bits` mentioned earlier, although that case does not represent a fixed capacity. I suppose we could split `ConstBits` and runtime `Bits`.

tarcieri:
@cuviper yeah, the use cases are similar to `BigUint::bits`, although it would be nice to be able to compute the value in a data-independent manner.

In the case of e.g. RSA the standard works around the significant bits vs bit capacity issue by ensuring the two are equal at key generation time, so either definition would work there.

src/bits.rs
```diff
@@ -0,0 +1,23 @@
+use core::{i128, i16, i32, i64, i8, isize};
+use core::{u128, u16, u32, u64, u8, usize};
+
+/// Numbers which have a specified number of bits.
+pub trait Bits {
```
tarcieri, line 5: What if you called this trait `Precision` instead? Then it could also have `bytes()`

tarcieri:
I assume many of the traits in this crate predate associated constants, but is there any thought to having both trait methods and associated constants? Perhaps in separate traits? (with a blanket impl linking them)

Associated constants would be nice for stack-allocated types, and methods for heap-allocated types.

tarcieri:
The nice thing about associated constants is they can be used in a `const fn` context on stable today, which as you noted won't be possible with trait methods until `const impl` is available.

dzmitry-lahoda:
Was rounding num-rational Ratio<u128> to Ratio<u64>. This would be nice trait to use in this case.

apps4uco:
Would a single trait with both a const and a function be a good idea?

Also I modified the macro to have Self::BITS instead of type::BITS

By the way my use case is for generic functions, when you want to shift by a certain amount depending on the width of the type in bits.

```rust 
pub trait Bits {
    const BITS: u32;
    fn bits() -> u32;
}

macro_rules! bits_impl {
    ( $($x:ident),* ) => (
        $(
            impl Bits for $x {
                const BITS: u32 = Self::BITS;

                fn bits() -> u32 { Self::BITS }
            }
        )*

    );
}

bits_impl!(
    i128, i16, i32, i64, i8, isize, u128, u16, u32, u64, u8, usize
);

```

tmvkrpxl0:
const trait is used by stdlib so I think it could be alternative. Similar strategy is used here: https://github.com/davidtwco/rfcs/blob/sized-hierarchy/text/3729-sized-hierarchy.md
Where things that depend on runtime information like SIMD vector width is determined in runtime for `Sized`

PR #275, opened 2023-07-03
Clarify `mul_add` Docs

rs017991:
There is some great documentation on the `MulAdd` trait itself, but in cases where you only see the method doc (like hovering in an IDE), it's not clear what operation will take place.

Sometimes I remember what role each of the three numbers has, and even where the parens are, but either way it breaks my concentration having to drill into the impl to find the trait doc to double-check my assumptions.

Short of copy-pasting the excellent `MulAdd` trait doc everywhere, I think that just including the pseudocode describing the operation (as I've done in this PR) will go a long way in clarifying what it does.

### DIFF
--- a/src/ops/mul_add.rs
+++ b/src/ops/mul_add.rs
@@ -28,9 +28,9 @@ pub trait MulAdd<A = Self, B = Self> {
     fn mul_add(self, a: A, b: B) -> Self::Output;
 }
 
-/// The fused multiply-add assignment operation `self = (self * a) + b`
+/// The fused multiply-add assignment operation `*self = (*self * a) + b`
 pub trait MulAddAssign<A = Self, B = Self> {
-    /// Performs the fused multiply-add assignment operation `self = (self * a) + b`
+    /// Performs the fused multiply-add assignment operation `*self = (*self * a) + b`
     fn mul_add_assign(&mut self, a: A, b: B);
 }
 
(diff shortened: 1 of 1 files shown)


src/ops/mul_add.rs
```diff
@@ -24,13 +24,13 @@ pub trait MulAdd<A = Self, B = Self> {
...
-    /// Performs the fused multiply-add operation.
+    /// Performs the fused multiply-add operation `(self * a) + b`
     fn mul_add(self, a: A, b: B) -> Self::Output;
 }
 
-/// The fused multiply-add assignment operation.
+/// The fused multiply-add assignment operation `self = (self * a) + b`
 pub trait MulAddAssign<A = Self, B = Self> {
-    /// Performs the fused multiply-add operation.
+    /// Performs the fused multiply-add assignment operation `self = (self * a) + b`
```
cuviper, line 33: Since `self` is a reference, this is really `*self = (*self * a) + b`, which you can also see in the impls below. I hope that doesn't look confusing, but if so we should probably use more words to describe it.

rs017991: Yeah I thought that pseudocode representing the mathematical operation taking place might read better, but if the other way is preferred then I'm cool with that.

cuviper:
Thanks!

bors r+

PR #276, opened 2023-07-04
Make Zero and One const traits

sigurd4:
Hi, i'm a a big fan of this library, and use it often. I'm currently working on some libraries using compile-time traits and functions, and figured it would be very convenient if the zero and one traits in your library are const traits.

Surely these traits should be const (at some point), however this requires enabling experimental language features, which i'm not sure is desireable at this point. I can understand if you guys want to wait with that until const traits are a stable rust language feature.

### DIFF
--- a/src/identities.rs
+++ b/src/identities.rs
@@ -1,3 +1,4 @@
+use core::marker::Destruct;
 use core::num::Wrapping;
 use core::ops::{Add, Mul};
 
@@ -9,6 +10,7 @@ use core::ops::{Add, Mul};
 /// a + 0 = a       ∀ a ∈ Self
 /// 0 + a = a       ∀ a ∈ Self
 /// ```
+#[const_trait]
 pub trait Zero: Sized + Add<Self, Output = Self> {
     /// Returns the additive identity element of `Self`, `0`.
     /// # Purity
@@ -20,7 +22,8 @@ pub trait Zero: Sized + Add<Self, Output = Self> {
     fn zero() -> Self;
 
     /// Sets `self` to the additive identity element of `Self`, `0`.
-    fn set_zero(&mut self) {
+    fn set_zero(&mut self)
+    where Self: ~const Destruct {
         *self = Zero::zero();
     }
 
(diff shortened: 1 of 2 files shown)


cuviper:
> Surely these traits should be const (at some point),

_Optionally_, yes, but some types can't be const because they involve allocation, like `BigUint::one()`.

> however this requires enabling experimental language features, which i'm not sure is desireable at this point. I can understand if you guys want to wait with that until const traits are a stable rust language feature.

Yeah, I don't want to deal with unstable features in general. Definitely not unconditional, but even putting it behind a "nightly" cargo feature would be work to keep up, and especially here given that const traits are in flux lately.

tarcieri:
Could there be separate traits with associated constants for zero and one? (and potentially, a blanket impl of `Zero` or `One` for those hypothetical new traits)

That could be a purely additive change that would also have the benefit of working on stable Rust.

haennes:
would be maintaining a "nightly" branch be an option #283

cuviper:
> would be maintaining a "nightly" branch be an option

I'm not interested in maintaining that... but as a branch, anyone could do this in their own fork.

haennes:
> > would be maintaining a "nightly" branch be an option
> 
> I'm not interested in maintaining that... but as a branch, anyone could do this in their own fork.

Ok. Hopefully const generics become stable soon. If/When my need for this arises again I will create a fork. Once I do this I will mention this issue

tarcieri:
I think it'd be possible to get similar benefits in a purely additive manner through a separate trait which defines them as consts, and could be linked to the existing traits via blanket impls:

```rust
pub trait ZeroConst {
    type ZERO: Self;
}

impl<T> Zero for T
where
    T: Add<Self, Output = Self> + Eq + ZeroConst
{
    fn zero() -> Self {
        Self::ZERO
    }

    fn is_zero(&self) -> bool {
        self.eq(&Self::ZERO)
    }
}
```

brundonsmith:
Just chiming in to say I'd love this feature too, and doing it as separate traits would be perfectly acceptable

tarcieri:
I opened a PR which adds `ZeroConstant` and `OneConstant` traits here: https://github.com/rust-num/num-traits/pull/303

tarcieri:
With #303 merged I think this can be closed?

PR #277, opened 2023-07-14
Add `--generate-link-to-definition` option when building on docs.rs

GuillaumeGomez:
This option generates links in source code pages, allowing to jump to definition and to jump back to doc. It makes browsing the source code pages much nicer. You can see it in action [here](https://doc.rust-lang.org/stable/nightly-rustc/src/rustc_middle/lib.rs.html#90).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ rust-version = "1.31"
 
 [package.metadata.docs.rs]
 features = ["std"]
+rustdoc-args = ["--generate-link-to-definition"]
 
 [dependencies]
 libm = { version = "0.2.0", optional = true }


cuviper:
Are other crates adopting it in this way yet?

Doesn't this also require `-Z unstable-options`?

GuillaumeGomez:
`docs.rs` uses only nightly and already uses `-Z unstable-options`. `dtolnay` crates are using it, some from `burnsushi`, and a few more as well. There are two goals for me: gather some documentation readers feedback and get some maintainers feedback as well.

And as a side-goal: having better docs to browse when reading source code pages. :)

cuviper:
OK, thanks!

bors r+

Issue #245, opened 2022-08-13
[feature request] Remove `Copy` bound from `Real`


CraftSpider:
Basically the reverse request of #243 - Currently, `Real` implies `Copy`, which restricts it from being implemented for real-value types that use non-copy internals (for example, a type that represents values as whole fractions backed by a `BigInt`). The `Copy` bound doesn't seem to be actually useful to the trait, and more like a holdover from when it was split out from `Float`.

This would be a breaking change unfortunately.

CraftSpider:
Just an added thought - If this change is made, it may be a good time to add `Bounded` as a super-trait and remove the `min_value` and `max_value` methods, as they're documented to do the same things

PR #284, opened 2023-08-18
Remove the `Copy` trait

LucaCiucci:
This was requested in #245. I just removed the `Copy` bound from the `Real` and replaced with `Clone` in `Float`.
This would be a breaking change.


nsabovic:
I for one would love that as it would enable quite a bit of use cases for non-copyable types. 👍

LucaCiucci:
Yeah, I think this would be really useful for working with automatic differentiation or arbitrary precision types. AFAIK currently the only way to use rust-num with these types is using a `[patch]`, but this is not feasible in many scenarios where you would have to patch many o crates to make things work.

Issue #285, opened 2023-09-18
`cargo doc` fails if they run it with nightly Rust


A1-Triard:
STR: run `cargo +nightly doc`.
ER: no errors.
AR:
```
error[E0658]: custom classes in code blocks are unstable
  --> src/identities.rs:4:1
   |
4  | / /// Defines an additive identity element for `Self`.
5  | | ///
6  | | /// # Laws
7  | | ///
...  |
10 | | /// 0 + a = a       ∀ a ∈ Self
11 | | /// ```
   | |_______^
   |
   = note: see issue #79483 <https://github.com/rust-lang/rust/issues/79483> for more information
   = help: add `#![feature(custom_code_classes_in_docs)]` to the crate attributes to enable
   = note: found these custom classes: class=text

error[E0658]: custom classes in code blocks are unstable
  --> src/identities.rs:80:1
   |
80 | / /// Defines a multiplicative identity element for `Self`.
81 | | ///
82 | | /// # Laws
83 | | ///
...  |
86 | | /// 1 * a = a       ∀ a ∈ Self
87 | | /// ```
   | |_______^
   |
   = note: see issue #79483 <https://github.com/rust-lang/rust/issues/79483> for more information
   = help: add `#![feature(custom_code_classes_in_docs)]` to the crate attributes to enable
   = note: found these custom classes: class=text
```

cuviper:
I think this is a rustdoc regression, as there was no error from nightly a week ago.

robamu:
I think  this might be due to https://github.com/rust-lang/rust/pull/110800

PR #286, opened 2023-09-18
Bugfix for text codeblock in documentation.

robamu:
Fixes https://github.com/rust-num/num-traits/issues/285 .
From what I have seen, the local output generated with `cargo +nightly doc --open` (rustc v1.74.0 nightly 203c57dbe) does not look differently than the most current one found here: https://docs.rs/num-traits/latest/num_traits/identities/trait.One.html

Propably related to https://github.com/rust-lang/rust/pull/110800 ? If it is, then the issue might be fixed soon and this PR should be ignored..

### DIFF
--- a/src/identities.rs
+++ b/src/identities.rs
@@ -5,7 +5,7 @@ use core::ops::{Add, Mul};
 ///
 /// # Laws
 ///
-/// ```{.text}
+/// ```text
 /// a + 0 = a       ∀ a ∈ Self
 /// 0 + a = a       ∀ a ∈ Self
 /// ```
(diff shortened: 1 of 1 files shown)


cuviper:
Looks like rust-lang/rust#115914 just relaxed that to a warning, but that's still annoying. Since your change has the right formatting without warning or error, even on older Rust, it's a reasonable change anyway. Thanks!

bors r+

kangalio:
Could a new patch release with this PR be pushed out soon? This presumably breaks CI workflow in many crates

OliverKillane:
@robamu could this be released as `num-traits-0.2.17`?

robamu:
I am not a maintainer of this crate. @cuviper is probably the better person to ask.

cuviper:
Published via #289.

PR #291, opened 2023-10-12
feat: add an extra method to (Checked)Euclid trait to get both quotien and rem

### LINKED ISSUE #290: add a quotien_and_rem_euclid method to Euclid trait
Sometimes you need both the quotient and the remainder.

With the current state of the trait you would have to call both `div_euclid` and `rem_euclid`.
But part of the computation done by each one of those methods can be shared.
See:
```rust
impl Euclid for f32 {
    #[inline]
    fn div_euclid(&self, v: &f32) -> f32 {
        let q = <f32 as crate::float::FloatCore>::trunc(self / v);
        if self % v < 0.0 {
            return if *v > 0.0 { q - 1.0 } else { q + 1.0 };
        }
        q
    }

    #[inline]
    fn rem_euclid(&self, v: &f32) -> f32 {
        let r = self % v;
        if r < 0.0 {
            r + <f32 as crate::float::FloatCore>::abs(*v)
        } else {
            r
        }
    }
}
```

can become
```rust
#[inline]
    fn div_and_rem_euclid(&self, v: &f32) -> (f32, f32) {
        let q = <f32 as crate::float::FloatCore>::trunc(self / v); 
        let r = self % v;
        if r < 0.0 {
            (if *v > 0.0 { q - 1.0 } else { q + 1.0 }, r + <f32 as crate::float::FloatCore>::abs(*v))
        } else {
            (q, r)
        }
    }
````
And you spare one computation of `self % v`.
This is a really small optimization for f32, but with the kind of big field elements I'm working with, it actually a pretty significant hit.
The method can be defined with a default impl:
```rust
fn div_and_rem_euclid(&self, v: &f32) -> (f32, f32) {
        (self.div_euclid(), self.rem_euclid())
}
```

Which has not optimization nor drawback for people using it, but can be overwritten in order to achieve better performances.

Wdyt?

tdelabro:
closes #290

### DIFF
--- a/src/ops/euclid.rs
+++ b/src/ops/euclid.rs
@@ -373,6 +373,8 @@ mod tests {
                     {
                         assert_eq!(CheckedEuclid::checked_div_euclid(&$t::min_value(), &-1), None);
                         assert_eq!(CheckedEuclid::checked_rem_euclid(&$t::min_value(), &-1), None);
+                        assert_eq!(CheckedEuclid::checked_div_euclid(&1, &0), None);
+                        assert_eq!(CheckedEuclid::checked_rem_euclid(&1, &0), None);
                     }
                 )+
             };
--- a/src/ops/euclid.rs
+++ b/src/ops/euclid.rs
@@ -50,7 +50,7 @@ pub trait Euclid: Sized + Div<Self, Output = Self> + Rem<Self, Output = Self> {
     /// Returns both the quotient and remainder from Euclidean division.
     ///
     /// By default, it internally calls both `Euclid::div_euclid` and `Euclid::rem_euclid`,
-    /// but it can be overidden in order to implement some optimization.
+    /// but it can be overridden in order to implement some optimization.
     ///
     /// # Examples
     ///
--- a/src/ops/euclid.rs
+++ b/src/ops/euclid.rs
@@ -47,11 +47,24 @@ pub trait Euclid: Sized + Div<Self, Output = Self> + Rem<Self, Output = Self> {
     /// ```
     fn rem_euclid(&self, v: &Self) -> Self;
 
-    /// Returns both the euclidian division quotien and remainer
+    /// Returns both the quotient and remainder from Euclidean division.
     ///
-    /// By default, it internaly calls both `Euclid::div_euclid` and `Euclid::rem_euclid`,
-    /// but it can be overwritten in order to implement some optimization.
-    fn div_with_rem_euclid(&self, v: &Self) -> (Self, Self) {
+    /// By default, it internally calls both `Euclid::div_euclid` and `Euclid::rem_euclid`,
+    /// but it can be overidden in order to implement some optimization.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use num_traits::Euclid;
+    /// let x = 5u8;
+    /// let y = 3u8;
+    ///
+    /// let div = Euclid::div_euclid(&x, &y);
+    /// let rem = Euclid::rem_euclid(&x, &y);
+    ///
+    /// assert_eq!((div, rem), Euclid::div_rem_euclid(&x, &y));
+    /// ```
+    fn div_rem_euclid(&self, v: &Self) -> (Self, Self) {
         (self.div_euclid(v), self.rem_euclid(v))
     }
 }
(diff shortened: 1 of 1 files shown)


src/ops/euclid.rs
```diff
@@ -46,6 +46,14 @@ pub trait Euclid: Sized + Div<Self, Output = Self> + Rem<Self, Output = Self> {
     /// assert_eq!(Euclid::rem_euclid(&-a, &-b), 1);
     /// ```
     fn rem_euclid(&self, v: &Self) -> Self;
+
+    /// Returns both the euclidian division quotien and remainer
+    ///
+    /// By default, it internaly calls both `Euclid::div_euclid` and `Euclid::rem_euclid`,
+    /// but it can be overwritten in order to implement some optimization.
```
cuviper, line 53:
```suggestion
    /// By default, it internally calls both `Euclid::div_euclid` and `Euclid::rem_euclid`,
    /// but it can be overridden in order to implement some optimization.
```

src/ops/euclid.rs
```diff
@@ -174,6 +182,14 @@ pub trait CheckedEuclid: Euclid {
     /// Finds the euclid remainder of dividing two numbers, checking for underflow, overflow and
     /// division by zero. If any of that happens, `None` is returned.
     fn checked_rem_euclid(&self, v: &Self) -> Option<Self>;
+
+    /// Returns both the checked euclidian division quotien and remainer
```
cuviper, line 186:
```suggestion
    /// Returns both the quotient and remainder from checked Euclidean division.
```

src/ops/euclid.rs
```diff
@@ -174,6 +182,14 @@ pub trait CheckedEuclid: Euclid {
...
+    ///
+    /// By default, it internaly calls both `Euclid::div_euclid` and `Euclid::rem_euclid`,
+    /// but it can be overwritten in order to implement some optimization.
```
cuviper, line 189:
```suggestion
    /// By default, it internally calls both `Euclid::checked_div_euclid` and `Euclid::checked_rem_euclid`,
    /// but it can be overridden in order to implement some optimization.
```

cuviper (changes requested):
I suggested some fixes for the documentation. I would also appreciate some tests, and doc-examples are an easy way to both demonstrate the feature and test it.

I think we should drop "with" from the name, just `div_rem_euclid` and `checked_div_rem_euclid`. This will be more similar to [`Integer::div_rem`](https://docs.rs/num-integer/latest/num_integer/trait.Integer.html#tymethod.div_rem).

src/ops/euclid.rs
```diff
@@ -46,6 +46,14 @@ pub trait Euclid: Sized + Div<Self, Output = Self> + Rem<Self, Output = Self> {
     /// assert_eq!(Euclid::rem_euclid(&-a, &-b), 1);
     /// ```
     fn rem_euclid(&self, v: &Self) -> Self;
+
+    /// Returns both the euclidian division quotien and remainer
```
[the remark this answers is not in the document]
tdelabro: done

src/ops/euclid.rs
tdelabro: done

src/ops/euclid.rs
tdelabro: donde

src/ops/euclid.rs
tdelabro: done

src/ops/euclid.rs
```diff
@@ -327,8 +373,6 @@ mod tests {
                     {
                         assert_eq!(CheckedEuclid::checked_div_euclid(&$t::min_value(), &-1), None);
                         assert_eq!(CheckedEuclid::checked_rem_euclid(&$t::min_value(), &-1), None);
-                        assert_eq!(CheckedEuclid::checked_div_euclid(&1, &0), None);
-                        assert_eq!(CheckedEuclid::checked_rem_euclid(&1, &0), None);
```
cuviper, line 331: Why was this removed?

tdelabro:
@cuviper Ready for review

src/ops/euclid.rs
```diff
@@ -46,6 +46,27 @@ pub trait Euclid: Sized + Div<Self, Output = Self> + Rem<Self, Output = Self> {
     /// assert_eq!(Euclid::rem_euclid(&-a, &-b), 1);
     /// ```
     fn rem_euclid(&self, v: &Self) -> Self;
+
+    /// Returns both the quotient and remainder from Euclidean division.
+    ///
+    /// By default, it internally calls both `Euclid::div_euclid` and `Euclid::rem_euclid`,
+    /// but it can be overidden in order to implement some optimization.
```
cuviper, line 53:
```suggestion
    /// but it can be overridden in order to implement some optimization.
```

cuviper:
Thanks!

bors r+

Issue #256, opened 2023-02-09
[feature request] total_cmp for num_traits::Float


felix-clark:
With `total_cmp()` being stabilized for `f32` and `f64`, could this be included in the floating point trait? It's quite a  convenient method for sorting generic floating point types.

cuviper:
Is there a generic way to write a default implementation? Otherwise, adding a trait method is a breaking change.

felix-clark:
To my knowledge there is no trivial built-in generic way.

cuviper:
Then I suppose this will need to be a standalone trait. Something like:

```
pub trait TotalOrd {
    fn total_cmp(&self, other: &Self) -> Ordering;
}
```

That could even have impls for integers. In theory, we could have a blanket impl for all `T: Ord`, but I think coherence would then disallow impls for `f32` and `f64` in case they add `Ord` in the future, even though we know they won't.

andrewjradcliffe:
Incidentally, [half](https://crates.io/crates/half) provides implementations of `total_cmp` for both `f16` and `bf16` which match the signature suggested by @cuviper. Likewise, [rug](https://crates.io/crates/rug)'s wrapper around MPFR provides a `total_cmp` implementation. In other words, it's little more than 5 lines of code for the respective maintainers.

As a separate constraint which could be imposed upon generic types, it would be useful in the rare numerical algorithms which necessitate sorting of floating point numbers (e.g. isotonic regression). `NaN` in such algorithms will of course produce garbage results, but using `total_cmp` makes the behavior reasonably transparent to the user -- in contrast to library writers making arbitrary decisions for how `NaN`s are handled (or not).

The inevitable bikeshedding: `TotalOrd` feels not quite as clean as one would like. `Ord` is already total. We could get around this by invoking the source of the ordering, i.e. name the trait `IeeeOrd` (or `IEEEOrd`?). While ugly to look at, it has the advantage of clarity.

cuviper:
We could further distinguish this from `PartialOrd`/`Ord` by spelling it out like `trait TotalOrder`, and include the standard library's documentation that this is "in accordance to the `totalOrder` predicate as defined in the IEEE 754 (2008 revision) floating point standard." (Or an approximation thereof for 3rd-party types.) We could also keep this in the `num_traits::float` module and *not* implement for integers as I had suggested before, so it remains scoped to float-like comparisons.

andrewjradcliffe:
[...]

PR #295, opened 2023-10-23
`TotalOrder` trait for floating point numbers

andrewjradcliffe:
Define an orthogonal trait which corresponds to the `totalOrder` predicate the IEEE 754 (2008 revision) floating point standard.

In order to maintain coherence, the bounds on `TotalOrder` should most likely be `TotalOrder: Float` (or `TotalOrder: FloatCore`).  Without type constraints, `TotalOrder` could be defined on, well, anything. Though slightly ugly, one way to deal with this is to define two traits, `TotalOrderCore: FloatCore` and `TotalOrder: Float`.  On the other hand, `Inv` has no such constraints (due to the possibility of a meaningful implementation on rational numbers). If the crate designers could weigh in on whether to introduce constraints, that would be helpful.

Resolves: [issue#256](https://github.com/rust-num/num-traits/issues/256)

### DIFF
--- a/src/float.rs
+++ b/src/float.rs
@@ -1,3 +1,4 @@
+use core::cmp::Ordering;
 use core::num::FpCategory;
 use core::ops::{Add, Div, Neg};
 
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -31,6 +31,7 @@ pub use crate::bounds::Bounded;
 #[cfg(any(feature = "std", feature = "libm"))]
 pub use crate::float::Float;
 pub use crate::float::FloatConst;
+pub use crate::float::TotalOrder;
```
cuviper, line 34: I think the best way to signal that this trait is only intended for floats is to *not* re-export it here, so you have to `use num_traits::float::TotalOrder`.

(And if people "abuse" it beyond that intent, it's not really our problem.)

src/float.rs
```diff
@@ -2210,6 +2210,74 @@ float_const_impl! {
...
+    /// check_lt(-0.0_f64, 0.0_f64);
+    /// ```
+    fn total_cmp(&self, other: &Self) -> std::cmp::Ordering;
+}
+macro_rules! totalorder_impl {
+    ($T:ident) => {
+        impl TotalOrder for $T {
+            #[inline]
+            fn total_cmp(&self, other: &Self) -> std::cmp::Ordering {
+                Self::total_cmp(&self, other)
```
cuviper, line 2273: We're going to need a fallback implementation for Rust < 1.62, or else this errors as unconditional recursion. I can add that myself if you're not sure how to go about that...

cuviper (changes requested):

src/lib.rs
andrewjradcliffe: Agree -- limiting exposure decreases probability of abuse.

src/float.rs
andrewjradcliffe: I don't know what sort of `cfg` tricks could be used to conditionally provide a fallback based on language version. Or maybe `cfg` is not the answer? I will learn from you.

cuviper: Ok, I've pushed that and a few other fixups -- please review in return!

andrewjradcliffe: Ah, the answer is `autocfg`... no surprise it has 200 million downloads. Very interesting.

Excellent idea to replace `std::cmp::Ordering` with `core::cmp::Ordering`, so that this PR is compatible with `no_std` by default.

Looks good to me.

Tangentially, thank you for the opportunity to contribute; I learned several useful things.

cuviper:
Thanks!

bors r+

TESLA-SATI:
#1554 - push

* build(deps): bump num-traits from 0.2.18 to 0.2.19

PR #298, opened 2023-11-13
Add constants Zero::ZERO and One::ONE

Sympatron:
Currently all trait methods are not usable in const contexts. Since zero is also the default value of all number types, this can be useful as a replacement for Default::default() in const contexts.

PS: Is there a reason that none of the traits provide constant versions of their values?


cuviper:
It's a breaking change, and allocated types like bigint can't support this. See also #276.

Issue #254, opened 2023-01-09
Why not `checked_pow` as a trait method


lemunozm:
Hi!

This is only a design question about the current implementation of `checked_pow` as a function. Why is it implemented as a function instead of using a trait method as the rest of the checked operations?

I'm implementing a wrapper over these functions, and I want to know if you found some issues that forced you to implement it as a plain function.

Thanks!

cuviper:
I think that's just historical -- it was added way back in rust-num/num#162, when `pow` was also still only a function, not yet a trait.

PR #300, opened 2023-11-25
Generic exp for pow and checked_pow

mtshr:
Due to the `exp` being `usize`, the current implementation has been inconsistent with core implementation of `pow` and `checked_pow` taking `u32` for `exp`. This has also inhibited implementation of some `Pow<u*>` for primitive integers.

This pull request generalizes `exp` to use unsigned primitive integer for the sake of the implementation of lacking `Pow<u*>` and more importantly, I guess, the consistency with core `pow` and `checked_pow` utilizing `u32`, for the use of architecture dependently sized type, namely `usize` here, for arithmetic function does not sound reasonable.

Although this change should be [a minor change](https://doc.rust-lang.org/cargo/reference/semver.html#fn-generalize-compatible) as `usize` does implements `PrimInt + Unsigned`, please note that the existing crate depending on the functions in question may come to be in need of adding some type notation, as {integer} is assumed to be `i32`.

```rust
num_traits::pow(7, 7);         // ng
num_traits::pow(7, 7usize);    // ok
```

### DIFF
--- a/src/pow.rs
+++ b/src/pow.rs
@@ -18,17 +18,24 @@ pub trait Pow<RHS> {
 }
 
 macro_rules! pow_impl {
+    (prim_int $t:ty) => {
+        pow_impl!($t, u8);
+        pow_impl!($t, u16);
+        pow_impl!($t, u32, u32, <$t>::pow);
+        pow_impl!($t, u64);
+        pow_impl!($t, u128);
+        pow_impl!($t, usize);
+    };
     ($t:ty) => {
         pow_impl!($t, u8);
+        pow_impl!($t, u16);
+        pow_impl!($t, u32);
+        pow_impl!($t, u64);
+        pow_impl!($t, u128);
         pow_impl!($t, usize);
-
-        // FIXME: these should be possible
-        // pow_impl!($t, u16);
-        // pow_impl!($t, u32);
-        // pow_impl!($t, u64);
     };
     ($t:ty, $rhs:ty) => {
-        pow_impl!($t, $rhs, usize, pow);
+        pow_impl!($t, $rhs, $rhs, pow);
     };
     ($t:ty, $rhs:ty, $desired_rhs:ty, $method:expr) => {
         impl Pow<$rhs> for $t {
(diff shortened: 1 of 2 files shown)


src/pow.rs
```diff
@@ -19,17 +18,24 @@ pub trait Pow<RHS> {
 }
 
 macro_rules! pow_impl {
+    (prim_int $t:ty) => {
+        pow_impl!($t, u8);
+        pow_impl!($t, u16);
+        pow_impl!($t, u32, u32, <$t>::pow);
```
mtshr, line 24: Utilizes the default implementation of pow for exp with u32

src/pow.rs
```diff
@@ -171,29 +127,33 @@ mod float_impls {
...
+/// assert_eq!(pow(2i8, 4u32), 16);
+/// assert_eq!(pow(6u8, 3u32), 216);
+/// assert_eq!(pow(0u8, 0u32), 1); // Be aware if this case affects you
 /// ```
 #[inline]
-pub fn pow<T: Clone + One + Mul<T, Output = T>>(mut base: T, mut exp: usize) -> T {
-    if exp == 0 {
+pub fn pow<T, U>(mut base: T, mut exp: U) -> T
+where
+    T: Clone + One,
```
mtshr, line 137: I removed `T: Mul<T, Output = T>` because `One` needs it.

cuviper:
It is an unfortunate historical difference, but I think the `{integer}` inference loss is probably not worth it when the more flexible `Pow<RHS>` already exists. Do you need an equivalent `CheckedPow`?

mtshr:
Thank you for taking a look.

> I think the `{integer}` inference loss is probably not worth it when the more flexible `Pow<RHS>` already exists.

I am sorry that my English ability is not sufficient, but am I right in interpreting you think it is too much to change `exp: usize` to general one and therefore are not in favor of the change of the pull request? What I give more importance here is actually not the fact that the user can implement their own `pow` with `Pow<RHS>`, but the reusability of the library's existing code and mitigating the concern about type size. After implementing `Clone` (and preferably `Copy` for the case I am describing at the moment considering the cost) + `One` for their own struct, the user will be able to just call `num_traits::pow` without writing their own `pow`. Also for example, for `Wrapping<T>` or something similar, the user can reasonably take exp >= 2^32, but always having to take `usize`'s actual size into account leads to implementing their own `pow` and in the end the user cannot benefit much from `pub pow`.

> Do you need an equivalent `CheckedPow`?

I know others may have a interest in one (ref #254) and I am not against the idea to add one. I consider making a pull request once the idea of generalizing exp is accepted, for I do not want to yield another FIXME.

https://github.com/rust-num/num-traits/blob/61d9a1ba1a52b3ebc7229c9a9dd5c99abc1c1e1a/src/pow.rs#L26-L29

I might have created an issue first. I apologize if discussing the idea here is quite inappropriate.

PR #303, opened 2023-12-14
Add `ConstZero` and `ConstOne` traits

tarcieri:
Adds traits which are alternatives to the more dynamic `Zero`/`One` traits which are useful for stack-allocated types where it's possible to define constant values for zero/one.

~~`ZeroConstant`~~ `ConstZero` bounds on `Zero` as a supertrait, and ~~`OneConstant`~~ `ConstOne` on `One`, allowing them to be used as drop-in replacements.

~~When a type also impls `PartialEq`, then `ZeroConstant` also provides a blanket impl of `Zero`, and likewise for `OneConstant`/`One`, making it simple for stack-allocated integers to impl these traits as an alternative to `Zero`/`One` while still remaining fully compatible.~~ (Edit: removed)

~~The internal impls of `Zero`/`One` on the numeric primitive types have been changed to use these traits, which should be a fully backwards-compatible change.~~ (Edit: removed)

Alternative to #276.

### DIFF
--- a/src/identities.rs
+++ b/src/identities.rs
@@ -38,20 +38,19 @@ pub trait ZeroConstant: Zero {
     const ZERO: Self;
 }
 
-impl<T: ZeroConstant + PartialEq> Zero for T {
-    #[inline(always)]
-    fn zero() -> T {
-        Self::ZERO
-    }
-
-    #[inline(always)]
-    fn is_zero(&self) -> bool {
-        self == &Self::ZERO
-    }
-}
-
 macro_rules! zero_impl {
     ($t:ty, $v:expr) => {
+        impl Zero for $t {
+            #[inline]
+            fn zero() -> $t {
+                $v
+            }
+            #[inline]
+            fn is_zero(&self) -> bool {
+                *self == $v
+            }
+        }
+
         impl ZeroConstant for $t {
             const ZERO: Self = $v;
         }
@@ -90,6 +89,13 @@ where
     fn zero() -> Self {
         Wrapping(T::zero())
     }
+}
+
+impl<T: ZeroConstant> ZeroConstant for Wrapping<T>
+where
+    Wrapping<T>: Add<Output = Wrapping<T>>,
+{
+    const ZERO: Self = Self(T::ZERO);
 }
 
 /// Defines a multiplicative identity element for `Self`.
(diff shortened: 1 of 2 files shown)


cuviper:
Do you also want to re-export these in the root? I think most people don't `use` the module hierarchy from here, which is honestly kind of a mess...

> When a type also impls `PartialEq`, then `ZeroConstant` also provides a blanket impl of `Zero`, and likewise for `OneConstant`/`One`, making it simple for stack-allocated integers to impl these traits as an alternative to `Zero`/`One` while still remaining fully compatible.

The problem is that these are not perfectly overlapping, especially with generics, like:

```rust
impl<T: Zero> Zero for Complex<T> { ... }
impl<T: ZeroConstant> ZeroConstant for Complex<T> { ... }
```

The first would include non-const types, but a blanket `impl<T: ZeroConstant> Zero for T` makes these conflict.

It's similar to Niko's old post about wanting `impl<T: Copy> Clone for T`:
https://smallcultfollowing.com/babysteps//blog/2016/09/24/intersection-impls/

tarcieri:
I have removed the blanket impls, and in their place impl'd `ZeroConstant` for `Wrapping` instead.

Sidebar: would names like `ZeroConst` and `OneConst` be better?

cuviper:
> Sidebar: would names like `ZeroConst` and `OneConst` be better?

Probably yes, or `ConstZero` and `ConstOne`. There's not a lot of precedent that I can find, but some unstable APIs in std do use an abbreviated prefix, like [`ConstParamTy`](https://doc.rust-lang.org/core/marker/trait.ConstParamTy.html#).

src/identities.rs
```diff
@@ -28,6 +28,16 @@ pub trait Zero: Sized + Add<Self, Output = Self> {
     fn is_zero(&self) -> bool;
 }
 
+/// Defines an associated constant representing the additive identity element
+/// for `Self`.
+///
+/// Types which impl both this trait and [`PartialEq`] will receive a blanket
+/// impl of the [`Zero`] trait.
```
tarcieri, line 35: This is no longer true.

```suggestion
```

src/identities.rs
```diff
@@ -115,6 +133,16 @@ pub trait One: Sized + Mul<Self, Output = Self> {
     }
 }
 
+/// Defines an associated constant representing the multiplicative identity
+/// element for `Self`.
+///
+/// Types which impl both this trait and [`PartialEq`] will receive a blanket
+/// impl of the [`One`] trait.
```
tarcieri, line 140: This is no longer true.

```suggestion
```

cuviper (approved):
Thanks! I rebased to pick up CI changes, and added a couple small fixes.

Issue #234, opened 2022-02-27
Add clamp(min, max) to Float


ncihnegn:
Since it is available for f32 and f64.

cuviper:
That's fine, as long as it's written with a generic default impl too, so it's not a breaking change.

PR #305, opened 2024-01-19
Implementing clamp for Float trait

michaelciraci:
Implementing `clamp` for Float per this issue: https://github.com/rust-num/num-traits/issues/234


cuviper:
We can't blindly forward to the inherent methods until Rust 1.50, but I'm planning an MSRV bump soon, stay tuned...

PR #310, opened 2024-02-08
Use namespaced-features to safely bump to MSRV 1.60

cuviper:
Earlier versions won't see the incompatible updates at all!

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,13 +9,7 @@ jobs:
     strategy:
       matrix:
         rust: [
-          1.31.0, # MSRV
-          1.35.0, # has_copysign
-          1.37.0, # has_reverse_bits
-          1.38.0, # has_div_euclid
-          1.44.0, # has_to_int_unchecked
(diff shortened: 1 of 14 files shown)


cuviper:
Note: I have not considered MSRV bumps to be breaking changes, but I have still been very conservative with them to avoid disrupting dependent crates. This bump is not meant to set any new policy, per se -- it's just that this particular feature of 1.60 offers an opportunity to raise MSRV without breaking anyone on older versions.

cuviper:
> Earlier versions won't see the incompatible updates at all!

Now that I have released 0.2.19 (#321), I find that this isn't completely true. Cargo 1.51 through 1.59 will ignore it thanks to the schema version in rust-lang/cargo#9161, backported in rust-lang/cargo#9196, but older Cargo doesn't know to look for that.

So, this bump isn't quite as "free" as I intended, but I think it will be OK. This MSRV 1.60 is still over 2 years old (2022-04-07), and the accidental schema-aware minimum of 1.51 is over 3 years old (2021-03-25). If someone really wants to use even older rustc, they can manually lock an older version of this crate.

Issue #213, opened 2021-04-08
Make FloatConst members be const


jturner314:
I'd like to propose the following breaking change (to be added to #47): make the members of `FloatConst` be either associated consts (my preference) or `const fn`s, so that they can be used in constants.

cuviper:
How would that be used? Last I checked, you couldn't have generic constraints (like `T: FloatConst`) in constant contexts at all. If you're using a specific type, it can (and might already) have those associated consts itself.

jturner314:
Unfortunately, `f32` and `f64` have very few inherent associated consts; most constants (including essential mathematical constants like `E` and `PI`) are in `std::f32::consts`/`std::f64::consts`. So, `FloatConst` is useful as an extension trait so that constants can be accessed through the names of the types, rather than through the separate `std::f32::consts`/`std::f64::consts` modules. For using `FloatConst` as an extension trait, it would be nice for the members of `FloatConst` to be `const`, for a couple of reasons:

1. For code which uses a type alias to switch between `f32` and `f64`, it's nice to have associated constants inherent to the type or as part of an extension trait, so that they can be easily accessed using the name of the type alias in constants:

   ```rust
   // Located in a common module/crate in the workspace.
   pub type Fxx = f64;

   // Located in a dependent module/crate.
   pub const ACTIONS: &[Action] = &[
       Action {
           speed: 20.,
           ang_vel: -Fxx::TAU / 15.,
       },
       Action {
           speed: 30.,
           ang_vel: 0.,
       },
       Action {
           speed: 20.,
           ang_vel: Fxx::TAU / 15.,
       },
   ];
   ```

2. When writing a module which uses *both* `f32` and `f64` explicitly for different purposes (i.e. some variables need to be higher precision, but others can be lower precision), it would be more concise to write `use num_traits::FloatConst` followed by `f32::PI` and `f64::PI`, instead of `use std::{f32, f64}` followed by `f32::consts::PI` and `f64::consts::PI`, in `const` contexts.

Both of these could be solved by moving the consts in `std::f32::consts`/`std::f64::consts` to be associated consts on `f32`/`f64`, but that seems unlikely to happen, so a trait like `FloatConst` is a nice alternative.

The only disadvantage I see to making the members of `FloatConst` be `const` is that it would prevent `FloatConst` from being implemented for non-`Copy` types.

cuviper:
[...]

PR #311, opened 2024-03-06
move to `Required Associated Constants`

BabaBert:
This merge request transitions the trait methods of associated constants to actual required constants.
It also adds `MANTISSA_DIGITS`, `DIGITS` and `RADIX` to the Float traits.
This would close https://github.com/rust-num/num-traits/issues/213

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "liveServer.settings.port": 5501
+}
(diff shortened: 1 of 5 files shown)


cuviper:
I'm not planning to process breaking changes anytime soon, so I'd rather leave this on the issue for now. A pull request will only bitrot over time. Thanks anyway!

BabaBert:
I fully understand that.
However, is it an option to add these associated on top of the current methods? That would be possible and not break any existing functionality.

tarcieri:
@BabaBert see #303 as an additive PR for traits with associated constants which was accepted

PR #319, opened 2024-04-12
add `Debug` trait to `pub trait Num`

### LINKED ISSUE #318: `struct` built on top of `pub trait Num` can't compile if `pub trait Num` doesn't impl `Debug`

hamirmahal:
fixes #318.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,7 +22,7 @@
 #[cfg(feature = "std")]
 extern crate std;
 
-use core::fmt;
+use core::fmt::{self, Debug};
 use core::num::Wrapping;
 use core::ops::{Add, Div, Mul, Rem, Sub};
 use core::ops::{AddAssign, DivAssign, MulAssign, RemAssign, SubAssign};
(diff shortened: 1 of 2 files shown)


cuviper:
Sorry, no, this would be a breaking change. I don't think `Debug` is fundamental to being numeric either.

hamirmahal:
I understand. Are you also against `pub trait Num` implementing `std::iter::Sum`?

```rust
#[test]
fn test_sum() {
    use num_traits::Num;
    pub struct LengthNArrayOfTypeT<const N: usize, T: Num>([T; N]);
    impl<const N: usize, T: Clone + Num> LengthNArrayOfTypeT<N, T> {
        fn dot(&self, v: &LengthNArrayOfTypeT<N, T>) -> T {
            self.0
                .iter()
                .zip(v.0.iter())
                .map(|(a, b)| a.clone() * b.clone())
                .sum()
        }
    }
}
```

could compile if `pub trait Num` implements `std::iter::Sum`, if I'm not mistaken.

cuviper:
Any added requirements are equally breaking changes.

PR #327, opened 2024-06-27
Add unit tests for trait method `float::Float::integer_decode`

mtilda:
### Motivation

While working on #326, I saw an opportunity to add unit test coverage for the trait method `float::Float::integer_decode`. I hope others find this helpful when troubleshooting this method.

### Changes

* Add unit test coverage for the trait method `Float::integer_decode` with 32-bit and 64-bit inputs.
* Update doc comments to clarify how to recover inputs to `Float::integer_decode`.

### DIFF
--- a/src/float.rs
+++ b/src/float.rs
@@ -2470,8 +2470,6 @@ mod tests {
             test_integer_decode(sign_f * f32::MAX, (0xffffff, 104, sign));
             test_integer_decode(sign_f * f32::INFINITY, (0x800000, 105, sign));
         }
-        // FIXME: Unclear if we should be able to recover NaN inputs
-        // test_integer_decode(f32::NAN, (0xc00000, 105, 1));
     }
 
     #[test]
@@ -2494,8 +2492,6 @@ mod tests {
             test_integer_decode(sign_f * f64::MAX, (0x1fffffffffffff, 971, sign));
             test_integer_decode(sign_f * f64::INFINITY, (0x10000000000000, 972, sign));
         }
-        // FIXME: Unclear if we should be able to recover NaN inputs
-        // test_integer_decode(f64::NAN, (0x18000000000000, 972, 1));
     }
 
     #[test]
--- a/src/float.rs
+++ b/src/float.rs
@@ -2456,6 +2456,7 @@ mod tests {
         for sign in [1, -1] {
             let sign_f = sign as f32;
             test_integer_decode(sign_f * 0.0__f32, (0x000000, -150, sign));
+            test_integer_decode(sign_f * 1.0e-40_f32, (0x022d84, -150, sign)); // subnormal (between 0 and MIN_POSITIVE)
             test_integer_decode(sign_f * f32::MIN_POSITIVE, (0x800000, -149, sign));
             test_integer_decode(sign_f * 0.25_f32, (0x800000, -25, sign));
             test_integer_decode(sign_f * 0.5__f32, (0x800000, -24, sign));
@@ -2478,6 +2479,7 @@ mod tests {
         for sign in [1, -1] {
             let sign_f = sign as f64;
             test_integer_decode(sign_f * 0.0__f64, (0x00000000000000, -1075, sign));
+            test_integer_decode(sign_f * 1.0e-308_f64, (0x0e61acf033d1a4, -1075, sign)); // subnormal (between 0 and MIN_POSITIVE)
             test_integer_decode(sign_f * f64::MIN_POSITIVE, (0x10000000000000, -1074, sign));
             test_integer_decode(sign_f * 0.25_f64, (0x10000000000000, -54, sign));
             test_integer_decode(sign_f * 0.5__f64, (0x10000000000000, -53, sign));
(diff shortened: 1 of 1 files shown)


mtilda:
@cuviper this is ready for review.

I'm especially interested to hear your thoughts about whether NaN values should be recoverable using the formula sign * mantissa * 2^exponent.

src/float.rs
```diff
@@ -2397,6 +2397,103 @@ mod tests {
...
+            test_integer_decode(sign_f * 2.5__f32, (0xa00000, -22, sign));
+            test_integer_decode(sign_f * 3____f32, (0xc00000, -22, sign));
+            test_integer_decode(sign_f * 4____f32, (0x800000, -21, sign));
+            test_integer_decode(sign_f * 5____f32, (0xa00000, -21, sign));
+            test_integer_decode(sign_f * 42___f32, (0xa80000, -18, sign));
+            test_integer_decode(sign_f * f32::MAX, (0xffffff, 104, sign));
+            test_integer_decode(sign_f * f32::INFINITY, (0x800000, 105, sign));
+        }
+        // FIXME: Unclear if we should be able to recover NaN inputs
+        // test_integer_decode(f32::NAN, (0xc00000, 105, 1));
```
mtilda, line 2470: This line panics with the message:

```
---- float::tests::integer_decode_f32 stdout ----
thread 'float::tests::integer_decode_f32' panicked at src/float.rs:2435:9:
absolute deviation must not exceed tolerance`
	    given:  NaN
	recovered:  +inf
	deviation:  NaN
	tolerance: <1e-6
```

src/float.rs
```diff
@@ -2397,6 +2397,103 @@ mod tests {
...
+            test_integer_decode(sign_f * 2.5__f64, (0x14000000000000, -51, sign));
+            test_integer_decode(sign_f * 3____f64, (0x18000000000000, -51, sign));
+            test_integer_decode(sign_f * 4____f64, (0x10000000000000, -50, sign));
+            test_integer_decode(sign_f * 5____f64, (0x14000000000000, -50, sign));
+            test_integer_decode(sign_f * 42___f64, (0x15000000000000, -47, sign));
+            test_integer_decode(sign_f * f64::MAX, (0x1fffffffffffff, 971, sign));
+            test_integer_decode(sign_f * f64::INFINITY, (0x10000000000000, 972, sign));
+        }
+        // FIXME: Unclear if we should be able to recover NaN inputs
+        // test_integer_decode(f64::NAN, (0x18000000000000, 972, 1));
```
mtilda, line 2494: This line panics with the message:

```
---- float::tests::integer_decode_f64 stdout ----
thread 'float::tests::integer_decode_f64' panicked at src/float.rs:2435:9:
absolute deviation must not exceed tolerance`
	    given:  NaN
	recovered:  +inf
	deviation:  NaN
	tolerance: <1e-6
```

cuviper:
> I'm especially interested to hear your thoughts about whether NaN values should be recoverable using the formula sign * mantissa * 2^exponent.

It's a good observation -- no, I don't think there's any way to recover a NaN, apart from recognizing the special exponent and taking a different branch. Even then it will be hard to match the exact NaN bits, signaling or not, etc.

Subnormals are also problematic, but those might be recoverable if you're more careful about `exp2` underflow. Something like this, but I haven't tested it:

```rust
if exponent >= MIN_EXP {
    sign * magnitude * exponent.exp2()
} else {
    // MIN_POSITIVE == 2^(MIN_EXP - 1)
    (sign * magnitude * MIN_POSITIVE) * (exponent - (MIN_EXP - 1)).exp2()
}
```

I'm ignoring type conversions there, and we don't have generic versions of those constants either. Even then, some hardware will flush subnormal arithmetic to zero anyway, so it still may not work.

So I think it's probably not worth the effort to document anything more complicated, and we can just say that the simple formula only works for zero/normal/infinite values (per `classify()`).

mtilda:
Thanks for the feedback!

---

> we can just say that the simple formula only works for zero/normal/infinite values (per classify()).

I just pushed a change to update the doc comment.

---

By the way, I just learned about Rust's [documentation testing](https://doc.rust-lang.org/rust-by-example/testing/doc_testing.html) feature 😅 I'm new to Rust, so thanks for being patient.

Should I move the tests I wrote to the method doc strings, or keep them in the `tests` module?

mtilda:
I just added test cases for subnormals. The recovery code seems to work (within tolerance) for those, but I think they are being treated as zero.

cuviper:
> By the way, I just learned about Rust's [documentation testing](https://doc.rust-lang.org/rust-by-example/testing/doc_testing.html) feature 😅 I'm new to Rust, so thanks for being patient.
> 
> Should I move the tests I wrote to the method doc strings, or keep them in the `tests` module?

Doc tests are nice, but we don't want to overwhelm the reader. I think yours are fine as unit tests.

> I just added test cases for subnormals. The recovery code seems to work (within tolerance) for those, but I think they are being treated as zero.

Yes, your `1e-6` tolerance is _much_ greater than the value in question, let alone the possible difference. The `MIN_POSITIVE` tests aren't going to be effective there either, and actually their exponent part will be less than `MIN_EXP` already due to the mantissa offset -- not sure if that will work out via subnormal temporaries or not.

PR #330, opened 2024-07-10
Two::two() and OneHalf::one_half()

sunsided:
This adds the traits `Two`, `ConstTwo`, as well as `OneHalf` and `ConstOneHalf` respectively. It's like `One`, just twice as good ... or half.

---

This may seem overkill at first glance, but I have found myself re-implementing

```rust
let two = T::one() + T::one();
let one_half = T::one() / two;
```

repeatedly when implementing digital filters that are either floating- or fixed-point. The need arises for example when differentiating nonlinear equations in order to provide their [Laplacians](https://en.wikipedia.org/wiki/Laplace_operator) for linear approximation. While a square can be implemented trivially (`a * a`), the differentiation of it (`2 * a`) causes the issue. 

---

I provided `OneHalf`/`ConstOneHalf` since they are dual to `Two`/`ConstTwo`. They could be derived, but that would require an additional `Div<T>` trait bound (whereas `Mul<T>` is already required by `One`). Given that multiplication is generally faster than division, having it at hand is useful.

### DIFF
--- a/src/identities.rs
+++ b/src/identities.rs
@@ -133,6 +133,44 @@ pub trait One: Sized + Mul<Self, Output = Self> {
     }
 }
 
+/// Defines twice the multiplicative identity element for `Self`.
+///
+/// # Laws
+///
+/// ```text
+/// a * 2 = a + a      ∀ a ∈ Self
+/// 2 * a = a + a      ∀ a ∈ Self
+/// ```
+pub trait Two: One {
+    /// Returns twice the multiplicative identity element of `Self`, `2`.
+    ///
+    /// # Purity
+    ///
+    /// This function should return the same result at all times regardless of
+    /// external mutable state, for example values stored in TLS or in
+    /// `static mut`s.
+    // This cannot be an associated constant, because of bignums.
+    fn two() -> Self;
+
+    /// Sets `self` to twice the multiplicative identity element of `Self`, `2`.
+    fn set_two(&mut self) {
+        *self = Two::two();
+    }
+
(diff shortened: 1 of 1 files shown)


sunsided:
Yeah, precisely why I said overkill. :) It's the same discussion in rust-lang about `FRAC_` constants.
In my case, it looks like

```rust
let one = T::one();
let two = one + one;
let x = vec.x * (one - two * (q2 * q2 + q3 * q3))
    + vec.y * two * (q1 * q2 - q0 * q3)
    + vec.z * two * (q1 * q3 + q0 * q2);

// and 
let y = vec.x * (q1 * q2 + q0 * q3)
    + vec.y * (one - two * (q1 * q1 + q3 * q3))
    + vec.z * two * (q2 * q3 - q0 * q1);

// and 
let z = vec.x * (q1 * q3 - q0 * q2)
    + vec.y * two * (q2 * q3 + q0 * q1)
    + vec.z * (one - two * (q1 * q1 + q2 * q2));
```

then in the next it's

```rust
let one = T::one();
let two = one + one;
let sinp = two * (w * y - z * x);
// TODO: If sin >= 1.0 || sin <= -1.0, clamp to +/- pi/2 instead
sinp.arcsin()
```

and then rinse and repeat. `PI` is a common factor, and you see the division by two in the note as well.
It's not like there is no solution, it's just getting repetitive and I figured others may have encountered similar situations.

_Personally_ I don't mind having these constants as much, but - as I said - I completely understand it's not exactly a popular standpoint. :)

cuviper:
Ok, I'm going to take the stance that this isn't really needed. Thanks anyway!

PR #331, opened 2024-07-11
Support Widening Mul

Lohann:


cuviper:
Thanks, although I would like to see the standard library API stabilized before we replicate it here.

PR #332, opened 2024-07-26
Fix checked operations documentation.

### LINKED ISSUE #13: Documentation for checked operations says "instead of wrapping around on underflow"
_From @kryptan on July 30, 2015 21:50_

Documentation for `CheckedAdd`, `CheckedDiv`, `CheckedMul`, `CheckedSub` says: Performs operation that returns `None` instead of wrapping around on underflow.

"instead of wrapping around on underflow" sounds like wrapping on overflow is the default behaviour but it is not.

P.S. there are traits for checked and saturating operations but no trait for wrapped operations, why? 

_Copied from original issue: rust-num/num#107_

ocstl:
Closes #13 . Remove the implication that wrapping is the default behaviour.

The documentation mentions "underflow" (for example, in `CheckedSub`), which I left as is. But Rust's documentation tends to refer to "overflow" exclusively (see [i32::checked_sub](https://doc.rust-lang.org/std/primitive.i32.html#method.checked_sub)).

I've recently been made aware that overflow is probably the correct term in both cases for integers (positive and negative), while underflow is strictly reserved for floating points. Either way, it's not a big deal and I'm not bothered by it, as the meaning is clear, but I thought I should mention it.

### DIFF
--- a/src/ops/checked.rs
+++ b/src/ops/checked.rs
@@ -1,7 +1,6 @@
 use core::ops::{Add, Div, Mul, Rem, Shl, Shr, Sub};
 
-/// Performs addition that returns `None` instead of wrapping around on
-/// overflow.
+/// Performs addition, returning `None` if overflow occurred.
 pub trait CheckedAdd: Sized + Add<Self, Output = Self> {
     /// Adds two numbers, checking for overflow. If overflow happens, `None` is
     /// returned.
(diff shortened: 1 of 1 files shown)


cuviper:
Let's change "underflow" -- the reference calls it "overflow" for results less than the minimum:
https://doc.rust-lang.org/reference/expressions/operator-expr.html#overflow

... and "underflow" isn't found in the reference at all!

ocstl:
I guess the reference settles that!

I also missed the `CheckedEuclid` methods in the first commit, but they are now fixed as well.

cuviper:
Thanks!

Issue #98, opened 2019-01-17
Float::integer_decode cannot be implemented for quadruple precision floating types.


jkarns275:
This is the definition of `integer_decode`:

```rust
fn integer_decode(self) -> (u64, i16, i8)
```

The 64 bits allocated for the mantissa is too small, 112 bits would be required to store the mantissa of a IEEE 754 quadruple-precision float. The exponent requires 15 bits, so i16 is enough, and naturally i8 is enough for the sign.

I ran into this issue while working on bindings to the GCC __float128 type (aka `libquadmath`). Would changing the u64 meant for the mantissa to a u128 be too much of a breaking change?  If not, I will implement it in the next couple of days.

cuviper:
It would definitely be a breaking change to change the current signature.

But we could add an alternate like `fn integer_decode_ex(self) -> (u128, i16, i8)` (insert name bikeshedding here), with a default implementation that just extends from `integer_decode`. Your `__float128` implementation would have to be lossy on the non-ex version, I guess.

I think you actually need `i32` for negative exponents, since our definition as `mantissa * 2^exponent` forces an extra mantissa shift. e.g. `integer_decode_64` does this:

```rust
    // Exponent bias + mantissa shift
    exponent -= 1023 + 52;
```

Another option is to add full flexibility in a separate trait, something like:

```rust
trait FloatDecode {
    type Mantissa;
    type Exponent;
    type Sign;

    fn decode(&self) -> (Self::Mantissa, Self::Exponent, Self::Sign);
    // perhaps individual `decode_mantissa` etc. too
}
```

This would allow arbitrary types, like IEEE 754 octuple precision or a bigint-based float.

jkarns275:
I like the idea of a FloatDecode trait, but it would effectively deprecate the current integer_decode method (which may be perfectly fine). I will work on an impl of the FloatDecode trait in the next couple of days, and ask for some feedback in this thread. 

Also, I don't quite understand why an i32 is needed for the exponent. The way I'm thinking about it is that the exponent range for f128 is `-16382..=16383`, which neatly fits in an i16 with room to spare.

cuviper:
The minimum exponent is only -16382 when you're interpreting the mantissa as 1.xxxx... But since we're returning the mantissa as an integral 1xxxx... value, the exponent needs a further negative bias.

[...]

PR #333, opened 2024-08-26
Add f16, f128

usamoi:
1. `integer_decode` is implemented for `f128` by truncating the radix to its 64 most significant bits. See #98.
2. Because of precision limitations, most implementation are inaccurate under `no_std` and `libm`, and a test fail because of this.

### DIFF
--- a/src/float.rs
+++ b/src/float.rs
@@ -2,6 +2,8 @@ use core::cmp::Ordering;
 use core::num::FpCategory;
 use core::ops::{Add, Div, Neg};
 
+#[cfg(has_f128)]
+use core::f128;
 #[cfg(has_f16)]
 use core::f16;
 use core::f32;
@@ -779,6 +781,10 @@ pub trait FloatCore: Num + NumCast + Neg<Output = Self> + PartialOrd + Copy {
     /// Returns the mantissa, base 2 exponent, and sign as integers, respectively.
     /// The original number can be recovered by `sign * mantissa * 2 ^ exponent`.
     ///
+    /// Note: For `f128`, this function is lossy. The original `f128` value cannot
+    /// be exactly recovered by `sign * mantissa * 2 ^ exponent` due to the precision
+    /// limitation of 64-bit mantissa.
+    ///
     /// # Examples
     ///
     /// ```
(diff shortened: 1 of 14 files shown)


src/lib.rs
```diff
@@ -17,6 +17,8 @@
 #![doc(html_root_url = "https://docs.rs/num-traits/0.2")]
 #![deny(unconditional_recursion)]
 #![no_std]
+#![cfg_attr(feature = "f16", feature(f16))]
+#![cfg_attr(feature = "f128", feature(f128))]
```
cuviper, line 21: I'm not keen on maintaining for unstable features in this crate, so we're probably going to be blocked here until the language and standard library stabilize these new floats.

At that point, I think we'll probably want to have detection in the build script to enable it automatically, and Cargo features can force that on -- like we used to for `has_i128`.

src/float.rs
```diff
@@ -2074,10 +2197,22 @@ fn integer_decode_f64(f: f64) -> (u64, i16, i8) {
     (mantissa, exponent, sign)
 }
 
+// FIXME: implement it
+#[cfg(feature = "f128")]
+fn integer_decode_f128(_f: f128) -> (u64, i16, i8) {
+    unimplemented!("radix part of `f128` is wider than `u64`")
```
cuviper, line 2203: We _could_ deal with this lossily by truncating the radix to its 64 most significant bits. That would need careful documentation, but I think it's better than just panicking, and we can still explore a fuller alternative per #98.

src/float.rs
```diff
@@ -777,6 +781,11 @@ pub trait FloatCore: Num + NumCast + Neg<Output = Self> + PartialOrd + Copy {
     /// Returns the mantissa, base 2 exponent, and sign as integers, respectively.
     /// The original number can be recovered by `sign * mantissa * 2 ^ exponent`.
     ///
+    /// This function is implemented for `f128`, however, lossily. The mantissa part
+    /// of `f128` is 113 bits, which does not fit in 64 bits. This function truncates
+    /// the radix to its 64 most significant bits, and the original number can be
+    /// estimated by `sign * mantissa * 2 ^ 49 * 2 ^ exponent`.
```
cuviper, line 787: Can't we add that 49 into the returned exponent too?

usamoi: Fixed.

scimind2460:
[rust-lang/libs-team#707](https://github.com/rust-lang/libs-team/issues/707) would eventually rely on this (as per my implementation) so could we expedite support?

cuviper:
> [rust-lang/libs-team#707](https://github.com/rust-lang/libs-team/issues/707) would eventually rely on this (as per my implementation) so could we expedite support?

I would not expect the standard library to start using `num-traits`, *especially* because that new `Complex` type should probably be in `core`, and you can't build `num-traits` as a dependency before that! I would expect a pseudo-internal trait instead, more like the setup of `struct NonZero<T> where T: ZeroablePrimitive`.

(My reservations remain about unstable features as well.)

folkertdev:
Trying this again, though I think the reservations about unstable features are valid: 

We're [aiming](https://rust-lang.github.io/rust-project-goals/2026/stabilizing-f16.html) to stabilize `f16` this year, and as far as we know support is complete when using LLVM 22 (which nightly rust now does).

For testing `portable_simd` it would be helpful if this crate supported `f16`. Specifically, the `portable_simd` crate uses `float-cmp`, which in turn uses `num-traits`. We'd like to test with `f16` vectors, but then that requires support in both of those dependencies.

I'd propose to use cargo features, similar to what I implemented for `proptest` https://github.com/proptest-rs/proptest/pull/638, and I think similar to `has_i128` that this crate already has (but I'm guessing the msrv got bumped high enough that `i128` is always available, so that feature is now no longer needed).

The `f128` type still has lots of issues, I don't think it makes sense to add it at this point in time.

cuviper:
For testing purposes, you could just `[patch.crates-io.num-traits]` to an `f16`-enabled fork, even on a local path.

Issue #274, opened 2023-06-30
`CheckedAdd` should probably not extend `Add`


popematt:
I'd like to contribute a PR to implement `CheckedAdd` for the `NonZero___` types in [`std::num`](https://doc.rust-lang.org/std/num/index.html), but it is not possible since they do not implement `core::ops::Add` (for good reason). The way to unblock my desired change is to make `CheckedAdd` not extend `Add`.

I suspect that removing the relationship between `CheckedAdd` and `Add` might be a breaking change, so I figured I'd propose the idea before I start a PR for it.

The presence of an unchecked add implies that a checked add is possible, but a checked add does not imply that an unchecked add is possible. I think that, if anything, `Add` should extend `CheckedAdd` (though I realize the practical difficulties of that). Given the relationship between them, I wonder if the real "right" solution also involves getting checked operations into `core::ops`. If it is, then I am willing to pursue that course of action instead of making changes to `num-traits`.

cuviper:
> I suspect that removing the relationship between `CheckedAdd` and `Add` might be a breaking change,

Yes, that is a breaking change, sorry.

> I wonder if the real "right" solution also involves getting checked operations into `core::ops`.

Those traits are all related to built-in language operators, so you can `impl Add` to get a working `+` operator on your own types. It's not impossible to think of having core traits for more numeric methods, but there's not any current precedent for that. Most of these `num` traits were in the standard library before Rust 1.0, and were kicked out to live here instead.

popematt:
> Most of these num traits were in the standard library before Rust 1.0, and were kicked out to live here instead.

Do you know where I can find any of the history of that decision?

cuviper:
I found the commit history, but not much discussion. For this particular trait:
* It was removed with its methods transferred to `trait Int` in https://github.com/rust-lang/rust/commit/e51cc089da7f5a067d348ee48f494c5bca662f95
* `trait Int` was deprecated in https://github.com/rust-lang/rust/commit/232424d9952700682373ccf2d578109f401ff023
* `trait Int` was removed in https://github.com/rust-lang/rust/commit/eeb94886adccb3f13003f92f117115d17846ce1f
  * and backported before 1.0.0 in https://github.com/rust-lang/rust/commit/0372484acc0270585479caf8cdccc6eb0106d8be

dzmitry-lahoda:
Can it be done like:
- detect major version change in this crate, so breaking change will be enabled automatically on toml update. 
[...]

PR #334, opened 2024-08-29
Implement Bounded, ToPrimitive, NumCast, FromPrimitive for NonZero*

acrrd:
I have a use case where I convert to and from primitive, and I need to involve `NonZero*` types as well.
Currently, I have a function that uses these traits for primitive types, and I have additional functions for each `NonZero*` that I need. It would be nice to get rid of them and have only one.

I can create on PR per trait implementation if it's preferred. This PR has one commit per trait implementation.

### DIFF
--- a/src/bounds.rs
+++ b/src/bounds.rs
@@ -1,4 +1,8 @@
 use core::num::Wrapping;
+use core::num::{
+    NonZeroI128, NonZeroI16, NonZeroI32, NonZeroI64, NonZeroI8, NonZeroIsize, NonZeroU128,
+    NonZeroU16, NonZeroU32, NonZeroU64, NonZeroU8, NonZeroUsize,
+};
 use core::{f32, f64};
 use core::{i128, i16, i32, i64, i8, isize};
 use core::{u128, u16, u32, u64, u8, usize};
@@ -67,6 +71,49 @@ bounded_impl!(i16, i16::MIN, i16::MAX);
 bounded_impl!(i32, i32::MIN, i32::MAX);
 bounded_impl!(i64, i64::MIN, i64::MAX);
 bounded_impl!(i128, i128::MIN, i128::MAX);
+
+macro_rules! bounded_impl_nonzero_const {
+    ($t:ty, $v:expr, $i:ident) => {
+        const $i: $t = {
+            let arr = [$v];
+            let idx = if $v != 0 { 0 } else { 1 };
+
+            unsafe { <$t>::new_unchecked(arr[idx]) }
+        };
+    };
+}
+
+macro_rules! bounded_impl_nonzero {
+    ($t:ty, $min:expr, $max:expr) => {
+        impl Bounded for $t {
+            #[inline]
+            fn min_value() -> $t {
+                bounded_impl_nonzero_const!($t, $min, MIN);
+                MIN
+            }
+
+            #[inline]
+            fn max_value() -> $t {
+                bounded_impl_nonzero_const!($t, $max, MAX);
+                MAX
+            }
+        }
+    };
+}
+
+bounded_impl_nonzero!(NonZeroUsize, 1, usize::MAX);
+bounded_impl_nonzero!(NonZeroU8, 1, u8::MAX);
+bounded_impl_nonzero!(NonZeroU16, 1, u16::MAX);
+bounded_impl_nonzero!(NonZeroU32, 1, u32::MAX);
+bounded_impl_nonzero!(NonZeroU64, 1, u64::MAX);
+bounded_impl_nonzero!(NonZeroU128, 1, u128::MAX);
+
+bounded_impl_nonzero!(NonZeroIsize, isize::MIN, isize::MAX);
+bounded_impl_nonzero!(NonZeroI8, i8::MIN, i8::MAX);
+bounded_impl_nonzero!(NonZeroI16, i16::MIN, i16::MAX);
+bounded_impl_nonzero!(NonZeroI32, i32::MIN, i32::MAX);
+bounded_impl_nonzero!(NonZeroI64, i64::MIN, i64::MAX);
+bounded_impl_nonzero!(NonZeroI128, i128::MIN, i128::MAX);
 
 impl<T: Bounded> Bounded for Wrapping<T> {
     fn min_value() -> Self {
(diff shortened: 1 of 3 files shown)


foolishell:
@acrrd 
Thank you so much for this pull request. If possible, could you also add support for `AsPrimitive`?

tarcieri:
Curious about https://doc.rust-lang.org/stable/core/num/struct.NonZero.html as well (might need an MSRV-preserving feature-gate)

acrrd:
> @acrrd Thank you so much for this pull request. If possible, could you also add support for `AsPrimitive`?

done

acrrd:
> Curious about https://doc.rust-lang.org/stable/core/num/struct.NonZero.html as well (might need an MSRV-preserving feature-gate)

This needs nightly, `NonZero` is only usable with types that implement [ZeroablePrimitive](https://doc.rust-lang.org/stable/core/num/trait.ZeroablePrimitive.html). This trait is usable only in nightly. 
When stabilized it will reduce the needed code to implement `NonZero`, but now is not usable.

tarcieri:
@acrrd though the `ZeroablePrimitive` trait itself is unstable, there are impls for it for the core integer types already which are usable on stable Rust:

https://doc.rust-lang.org/stable/core/num/trait.ZeroablePrimitive.html#implementors

Quite a bit of the functionality is already stable, as is noted in the feature gates in the rustdoc. Here's a small example:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9c4126921b16b781588732fd26d050a4

cuviper:
The old types are just aliases now, literally `type NonZeroI32 = NonZero<i32>;` etc. So implementing traits here on those aliases has exactly the same effect as implementing `NonZero<T>` for specific `T`.

cuviper:
I'm torn on `AsPrimitive` -- right now that is only implement for types where you could literally write `x as T` in non-generic settings. It's not sealed though, and the docs do mention how newtypes should behave too.

So if we add that, shouldn't we also do it for `Wrapping`? (and `Saturating`, if we get to that at all)

acrrd:
> I'm torn on `AsPrimitive` -- right now that is only implement for types where you could literally write `x as T` in non-generic settings. It's not sealed though, and the docs do mention how newtypes should behave too.

It was requested in #335, they have a use case for it. Let me know if you want to keepit.

> So if we add that, shouldn't we also do it for `Wrapping`? (and `Saturating`, if we get to that at all)

We can't. The trait requires `Add`,`Sub`, etc., which are not implemented for NonZero*. This is related to #274.

cuviper:
> > I'm torn on `AsPrimitive` -- [...]
> 
> It was requested in #335, they have a use case for it. Let me know if you want to keep it.

I'm going to respond there, because I think that use is also problematic, even aside from `NonZero`. Let's drop it here for now.

> > So if we add that, shouldn't we also do it for `Wrapping`? (and `Saturating`, if we get to that at all)
> 
> We can't. The trait requires `Add`,`Sub`, etc., which are not implemented for NonZero*. This is related to #274.

I wasn't talking about any other traits, only `impl<T, U> AsPrimitive<T> for core::num::Wrapping<U>`. But that deserves a separate PR, if we even decide that's worthwhile.

acrrd:
> I'm going to respond there, because I think that use is also problematic, even aside from `NonZero`. Let's drop it here for now.

done

> I wasn't talking about any other traits, only `impl<T, U> AsPrimitive<T> for core::num::Wrapping<U>`. But that deserves a separate PR, if we even decide that's worthwhile.

I see it now, sorry

marshrayms:
To me, a big part of the value of add-on crates like the `rust-num` collection is that they can help make up where the base language falls short. 

> I'm torn on `AsPrimitive` -- right now that is only implement for types where you could literally write `x as T` in non-generic settings. It's not sealed though, and the docs do mention how newtypes should behave too.

For example, the concept you mentioned of "types where you could literally write x as T" is not expressible using any combination of traits in the standard library.

> So if we add that, shouldn't we also do it for `Wrapping`? (and `Saturating`, if we get to that at all)

I think the difference is that `NonZero*` is a newtype which expresses an inherent property of the value itself. They are useful as type system constraints.

Whereas `Wrapping` and `Saturating` express nothing about the set of values they represent, they are simply used to select different sets of common functions via operator overloading. Purely syntactic convenience.

cuviper:
> To me, a big part of the value of add-on crates like the `rust-num` collection is that they can help make up where the base language falls short.
> 
> > I'm torn on `AsPrimitive` -- right now that is only implement for types where you could literally write `x as T` in non-generic settings. It's not sealed though, and the docs do mention how newtypes should behave too.
> 
> For example, the concept you mentioned of "types where you could literally write x as T" is not expressible using any combination of traits from the standard library.

This seems consistent to me -- `AsPrimitive` is currently filling that gap in the base language by making it expressible as a trait, though you only get `.as_()` instead of a **real** `as` in that generic code. We can choose to extend the domain of that trait as well, and maybe we should, but I think we should do that separately.

Anyway, let me give this PR an actual review...

src/cast.rs
```diff
@@ -266,6 +270,61 @@ impl_to_primitive_uint!(u32);
 impl_to_primitive_uint!(u64);
 impl_to_primitive_uint!(u128);
 
+macro_rules! impl_to_primitive_nonzero_to_method {
+    ($SrcT:ident : $( $(#[$cfg:meta])* fn $method:ident -> $DstT:ident ; )*) => {$(
+        #[inline]
+        $(#[$cfg])*
```
cuviper, line 276: I guess you copied the `cfg` from the other macros, but it's vestigial, from the days when `i128` was conditional. We definitely shouldn't propagate that, but could you clean up the others too?

src/cast.rs
```diff
@@ -266,6 +270,61 @@ impl_to_primitive_uint!(u32);
...
+                fn to_u8 -> u8;
+                fn to_u16 -> u16;
+                fn to_u32 -> u32;
+                fn to_u64 -> u64;
+                fn to_u128 -> u128;
+            }
+
+            #[inline]
+            fn to_f32(&self) -> Option<f32> {
+                Some(self.get() as f32)
```
cuviper, line 304: Why isn't this forwarding to the inner `to_f32` as well?

src/cast.rs
```diff
@@ -572,6 +631,85 @@ impl_from_primitive!(u128, to_u128);
 impl_from_primitive!(f32, to_f32);
 impl_from_primitive!(f64, to_f64);
 
+macro_rules! impl_from_primitive_nonzero {
+    ($T:ty, $to_ty:ident) => {
+        #[allow(deprecated)]
```
cuviper, line 636: I think this is also vestigial, way back from commit 03db5c9d1b7cc8625b507184ed3ecba9155b8cb8 when it was copied from its deprecated state in the standard library.

cuviper (changes requested):

src/bounds.rs
```diff
@@ -68,6 +72,49 @@ bounded_impl!(i32, i32::MIN, i32::MAX);
...
 bounded_impl!(i128, i128::MIN, i128::MAX);
 
+macro_rules! bounded_impl_nonzero_const {
+    ($t:ty, $v:expr, $i:ident) => {
+        const $i: $t = {
+            let arr = [$v];
+            let idx = if $v != 0 { 0 } else { 1 };
+
+            unsafe { <$t>::new_unchecked(arr[idx]) }
+        };
```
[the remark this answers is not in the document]
acrrd: Nice, done.

src/cast.rs
acrrd: I wasn't sure what it was so I just kept it. Removed here and from other macros

src/cast.rs
acrrd: Not sure, I think I copied from the other macro and didn't notice here I could put everything together.

src/cast.rs
acrrd: Done

acrrd:
@cuviper Do you think this is ok after the changes?

cuviper:
Thanks! And sorry for the slow review.

acrrd:
@cuviper do you plan to do a release soon with this? Is there any blocker I could help with?

Issue #215, opened 2021-05-19
Consider a trait for NonZero* integers


raphaelcohn:
I wanted a trait to identify `NonZero*` integers (eg `NonZeroU32`) and my first thought was this crate. I didn't seem a trait for them.

In my very specific use case, I actually need a trait that relates the zero and non-zero unsigned integers. Of course, I can trivially create one, but this seems more useful in this pseudo-standard Rust crate.

For example:-

```rust
trait NonZeroInteger
{
}

impl NonZeroInteger for NonZeroU32
{
}
```

and possibly

```rust
trait Integer
{
    type NonZeroInteger: NonZeroInteger
}
```

I suppose there's a debate to be had as to whether `NonNull` would also implement such a trait. I'd argue not - rather there should be a super trait of `NonZero` and a sub trait of `NonZeroInteger` - but that's quite arguable either way.

geeknoid:
See #85

tarcieri:
Also [`std::num::NonZero`](https://doc.rust-lang.org/std/num/struct.NonZero.html) provides a generic solution

Issue #85, opened 2018-09-15
Interest in generic nonzero and nonzeroable traits?


antifuchs:
Hey there, huge fan of the `num_traits` crate. I just needed traits to represent generic non-zero integer types (as opposed to the concrete `core::num::NonZeroU8` type), and figured this might be a good fit for this crate! In the meantime I made https://crates.io/crates/nonzero_ext, but would love to just fold it into this one (:

Happy to make the PR to add these! If you are interested, let me know!

PR #346, opened 2025-01-31
some set of generic based non zero support

dzmitry-lahoda:
TLDR;

- need new major release bump
- saturating sub works only for self minus self outputs self and requires sub, which is not the case generally
- etc

LONG:

Covers what can work for current stable non zero.

What can be done?
1. detect Rust/MSRV lang version and gate `NonZero<T>` since it was introduced. Easy, dtolnay doing that.
2. detect this crate version major bump, and change `base` traits relations. So we do not need `Add`/`Mul` on non zeros. add VERSION_NUMBER_UNSTABLE feature like. So as for now just add new base traits.
3. One. Seems reasonable to add, but it tells some MATH props on interface, I can reproduce it without need to `Mul`, with Checked only. 
4. Port tests  and some cast support from https://github.com/rust-num/num-traits/pull/334 .Differs from #334 by using generics.

I can do any version detection (Package/Rust) so can ensure it switched automatically.

Partially overlaps with next, but not same:

https://github.com/rust-num/num-traits/issues/85

https://github.com/rust-num/num-traits/issues/215

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ categories = ["algorithms", "science", "no-std"]
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/rust-num/num-traits"
 name = "num-traits"
-version = "0.2.19"
+version = "1.2.19"
 readme = "README.md"
 build = "build.rs"
 exclude = ["/ci/*", "/.github/*"]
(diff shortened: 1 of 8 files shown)


dzmitry-lahoda:
started my own https://github.com/dzmitry-lahoda/numz - so i do not support non Copy and allows `u32 * u64 = u96`

PR #365, opened 2025-08-13
Generalize Wrapping traits

rickwebiii:
I need to perform wrapping arithmetic where T and the right hand side are not the same type, which you cannot do with the existing Wrapping* traits. This change generalizes these traits so you can perform wrapping computation with a different RHS.

This will be a breaking change for anyone that implements Wrapping* traits, as they'll need to specify the associated type for the output.


rickwebiii:
I'm beginning to see how nasty this change will be to propagate...

Issue #377, opened 2026-04-16
function abs() is panicking on `min_value`


spremel:
The documentation states that `abs(::MIN)` returns `::MIN`. However, the current behavior is to panic on integer min value.
This is easily reproducible on the stable Rust playground.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=447346fe7ffc16a8c46eb694f30138ac

Thanks

crop2000:
Unrelated to num traits. (A max is smaller then a potential abs(min))

cuviper:
Oof, this is going **way** back. It's true that the documentation says that `MIN.abs()` returns `MIN`, and once upon a time this was always the case for the signed integer types. However, Rust 1.1 changed this:

https://doc.rust-lang.org/stable/releases.html#version-110-2015-06-25
> In accordance with Rust’s policy on arithmetic overflow `abs` now [panics on overflow when debug assertions are enabled](https://github.com/rust-lang/rust/pull/25441).

So for a very long time, the behavior has been that `MIN` panics with debug assertions (default in debug mode) and wraps back to `MIN` otherwise (as in release mode). We should update the documentation, though without being too prescriptive about how third-party types may handle this.

(This predates the `num-traits` crate itself, but these traits were split from the `num` crate which existed before Rust 1.0, originally split out from the standard library.)

spremel:
Thanks for clarifying this.

That's what I thought, I was mostly confused by the documentation that was saying that `MIN` is the result to expect. Since it is bit-wise arithmetically correct, that would have been an "acceptable" implementation of abs.

This can be closed then.

PR #380, opened 2026-08-03
Make Signed::abs match its documented behavior for ::MIN

tautschnig:
The `Signed` trait documentation states: "For signed integers, `::MIN` will be returned if the number is `::MIN`." The primitive implementations use unary negation, which panics with "attempt to negate with overflow" when overflow checks are enabled (debug builds):

```rust
let r = num_traits::abs(i32::MIN); // debug: panic; release: returns i32::MIN as documented
```

This PR switches the implementation to `wrapping_abs`, which implements the documented semantics in both build profiles, and adds a regression test (which runs with overflow checks enabled by default, so it would panic without the fix).

It also documents the overflow behavior of `abs_sub` for signed integers (e.g. `i32::MAX.abs_sub(&-1)` is not representable), which panics with overflow checks enabled and wraps otherwise; changing that behavior (e.g. saturating) seemed like a semantic decision better left to the maintainers, so this PR only documents it.

Found by running Kani's [autoharness](https://model-checking.github.io/kani/reference/experimental/autoharness.html) (model-checking/kani#3832) over the crate: the generated harness for `abs::<i32>` reported the negate-overflow panic reachable; reproduced with plain cargo (no Kani involved) before filing.


cuviper:
See https://github.com/rust-num/num-traits/issues/377#issuecomment-4263719585 -- I would rather update the documentation, so we keep matching the standard library behavior. In fact, I'm not sure why we're not just forwarding to the inherent `abs()` anyway.

PR #381, opened 2026-08-15
Introduce trait for (forced) total ordering of numbers

fogti:
Motivation: https://github.com/georust/rstar/pull/237#issuecomment-5303693983

### DIFF
--- a/src/float.rs
+++ b/src/float.rs
@@ -1,4 +1,3 @@
-use core::cmp::Ordering;
 use core::num::FpCategory;
 use core::ops::{Add, Div, Neg};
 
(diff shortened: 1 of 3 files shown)


cuviper:
We already have this, but for floats only:
https://docs.rs/num-traits/latest/num_traits/float/trait.TotalOrder.html

Can you say more about why you want the same trait for integers?

src/ops/total_ord.rs
```diff
@@ -0,0 +1,35 @@
...
+/// Returns the ordering between `self` and `other`.
+///
+/// For some types (particularly floating-point types), there exists a way to
+/// induce a canonical ordering on them, but as it might behave in unexpected
+/// ways, it isn't the default.
+pub trait TotalOrder {
+    fn total_cmp(&self, other: &Self) -> Ordering;
+}
+
+impl<T: crate::float::TotalOrder> TotalOrder for T {
```
cuviper: Oh, you're already using the other trait -- well, I would not want two by the same name. So if we really need integers too, then I'd move the trait and make `float::TotalOrder` a deprecated re-export.

fogti:
Mostly to be able to treat float numbers and integers uniformly in containers that don't really care about special float values and just need an arbitrary, but deterministic ordering for them, optimally to shift the burden away from the user of having to wrap float values in a wrapper that forwards `Ord::cmp` to `f_::total_cmp`.

In particularly, this applies to containers like priority queues or R-trees.

src/ops/total_order.rs
```diff
@@ -0,0 +1,121 @@
...
+macro_rules! totalorder_impl_via_ord_core {
+    ($($t:ty),*) => {
+        $(
+            impl TotalOrder for $t {
+                #[inline]
+                fn total_cmp(&self, other: &Self) -> Ordering {
+                    Self::cmp(self, other)
+                }
+            }
+            impl TotalOrder for NonZero<$t> {
```
michaelkirk, line 166: Consider also adding impls for [`Wrapping<$t>`](https://doc.rust-lang.org/std/num/struct.Wrapping.html)

src/ops/total_order.rs
```diff
@@ -0,0 +1,186 @@
...
+    i64, u64,
+    i128, u128,
+    isize, usize,
+    Ordering,
+);
+
+macro_rules! totalorder_impl_zeroable_via_ord_core {
+    ($($t:ty),* $(,)?) => {
+        $(
+            impl TotalOrder for NonZero<$t> {
```
fogti, line 166: Okay, only remaining problem afaik is that `NonZero` requires a bump of MSRV to 1.79

fogti:
I refactored this to minimize the amount of code generated by macros, by using generic trait impls instead.

src/ops/total_order.rs
```diff
@@ -0,0 +1,178 @@
...
+                fn total_cmp(&self, other: &Self) -> Ordering {
+                    <$t>::cmp(&self.get(), &other.get())
+                }
+            }
+        )*
+    }
+}
+
+totalorder_impl_zeroable_via_ord_core!(
+    char, i8, u8, i16, u16, i32, u32, i64, u64, i128, u128, isize, usize,
```
fogti, line 177: Weirdly enough `char: ZeroablePrimitive`, but `NonZeroChar` doesn't exist...

src/float.rs
```diff
@@ -2242,88 +2241,8 @@ float_const_impl! {
...
-                right ^= (((right >> ($bits - 1)) as $U) >> 1) as $I;
-
-                left.cmp(&right)
-            }
-        }
-    };
-}
-totalorder_impl!(f64, i64, u64, 64);
-totalorder_impl!(f32, i32, u32, 32);
+#[deprecated]
```
michaelkirk: It'd be nice to add a note to the deprecation which points at the new impl

PR #383, opened 2026-08-19
Add traits for distinguished 0 without `Add` and 1 without `Mul`

### LINKED ISSUE #364: Why does One require Mul?
While this is algebraically correct, implementations may not be able to provide `One` and `Mul` for the same types.

For example, I implemented polynomial quotient rings $(\mathbb{Z}/q\mathbb{Z})/(x^n+1)$ as a type `Poly` that implements `One`, but not `Mul` because naive polynomial multiplication would be inefficient and require a lot of code (division and extended Euclidean algorithm on polynomials). Instead, multiplication is done efficiently in the Number Theoretic Transform (NTT) domain, which is represented as a different type `PolyNtt` to ensure mathematical soundness. In this case I have to create my own trait `One` without the requirement.

mikwielgus:
This commit splits out functions defined in `Zero` into a trait called `PartialZero`. `Zero` is now `PartialZero`'s supertrait with additional requirement for `Add<Output = Self>`.

Same for `One` and `PartialOne`: `One` is now `PartialOne`'s supertrait with additional requirement for `Mul<Output = Self>`.

This is useful because some implementations have distinguished 0 and 1 elements but can't implement closed multiplication. There seems to be a considerable demand for that because issue #364 requesting that for `One` has 6 upvotes.

The names `PartialZero` and `PartialOne` are in analogy to `PartialEq` and `PartialOrd`. "partial zero" and "partial one" themselves are not mathematical terms. (if you disagree with this naming choice, let me know if you have a better name!)

Closes https://github.com/rust-num/num-traits/issues/364

### DIFF
--- a/src/float.rs
+++ b/src/float.rs
@@ -969,10 +969,10 @@ pub trait Float: Num + Copy + NumCast + PartialOrd + Neg<Output = Self> {
     /// Returns `-0.0`.
     ///
     /// ```
-    /// use num_traits::{Zero, Float};
+    /// use num_traits::{Float, PartialZero};
     ///
     /// let inf: f32 = Float::infinity();
-    /// let zero: f32 = Zero::zero();
+    /// let zero: f32 = f32::zero();
     /// let neg_zero: f32 = Float::neg_zero();
     ///
     /// assert_eq!(zero, neg_zero);
--- a/src/identities.rs
+++ b/src/identities.rs
@@ -4,16 +4,13 @@ use core::ops::{Add, Mul};
 #[cfg(has_num_saturating)]
 use core::num::Saturating;
 
-/// Defines an additive identity element for `Self`.
+/// Defines a distinguished `0` value for `Self`.
 ///
-/// # Laws
-///
-/// ```text
-/// a + 0 = a       ∀ a ∈ Self
-/// 0 + a = a       ∀ a ∈ Self
-/// ```
-pub trait Zero: Sized + Add<Self, Output = Self> {
-    /// Returns the additive identity element of `Self`, `0`.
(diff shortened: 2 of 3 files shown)


frozenspider:
Personally I think introducing `PartialZero`/`PartialOne` just complicates things for no real benefit.
I would prefer just dropping `Add`/`Mul` requirements on the `Zero`/`One` traits.

src/float.rs
```diff
@@ -969,10 +969,10 @@ pub trait Float: Num + Copy + NumCast + PartialOrd + Neg<Output = Self> {
     /// Returns `-0.0`.
     ///
     /// ```
-    /// use num_traits::{Zero, Float};
+    /// use num_traits::{Float, PartialZero};
     ///
     /// let inf: f32 = Float::infinity();
-    /// let zero: f32 = Zero::zero();
+    /// let zero: f32 = f32::zero();
```
fogti, line 975:
```suggestion
    /// let zero: f32 = PartialZero::zero();
```

src/identities.rs
```diff
@@ -22,15 +19,25 @@ pub trait Zero: Sized + Add<Self, Output = Self> {
...
+///
+/// ```text
+/// a + 0 = a       ∀ a ∈ Self
+/// 0 + a = a       ∀ a ∈ Self
+/// ```
+pub trait Zero: PartialZero + Add<Self, Output = Self> {}
+
 /// Defines an associated constant representing the additive identity element
 /// for `Self`.
 pub trait ConstZero: Zero {
```
fogti, line 43: Shouldn't the same be done for `ConstZero`, `ConstOne`?

cuviper:
We have no plans for making breaking changes. Let's keep further discussion on the issue.

